PR #1, opened 2020-08-20
add multi issuer support

alexmatson-da:
An attempt at adding support for multiple issuers. This changes the onboarding flow slightly - custodians must be onboarded before issuers.


dimitri-da (approved):
first PR! looks good! As you mentioned need to check the multi-custodian multi-issuer multi-exchange setup but that can come later

PR #5, opened 2020-09-01
Add broker role

alexmatson-da:


alexmatson-da:
Will close this for now -- the models around transfers have changed a bit since this attempt was tried. We can try implementing the Broker role again after some higher priority UI work is completed

PR #7, opened 2020-09-04
Clean up DAML app template for Marketplace UI

alexmatson-da:
This is part 1 of the work in converting the default DAML app UI template into what we need for the marketplace. This includes
 
- removing unnecessary files, 
- adding additional files that we'll need, 
- starting to include some custom site-wide CSS,
- and including the UI build in the Makefile target

The next PR will include the components for the Role Selector onboarding UX and the Investor view.


Sarahb-da (approved):
lgtm

PR #8, opened 2020-09-04
Add Role Selector, Investor Home & Wallet Views

alexmatson-da:
This PR begins adding some major components for the marketplace. 

- Login page
- Role selector page
- Investor home 
- Investor wallet

Along the way, I also fleshed out some reusable components based on the Semantic UI framework that will aid in developing further views. 

Some challenges:
One thing I had problems with was Semantic UI having built-in CSS that would have higher specificity over my own rules. That's why there are some pretty crazy CSS selectors in the code here.

I'm also wondering if we should switch to Sass. Being able to nest rules and mixins might help with the specificity issue, as well as help prevent accidental cross-styling between unrelated components.


ui/src/components/RoleSelectScreen.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const [ role, setRole ] = useState<MarketRole>();
+
+    const handleRoleClick = async (role: MarketRole, view: Mode) => {
+        setRole(role);
+        if (userSessions.length === 0) {
+            setLoading(true);
+            await ledger.create(UserSession, {
+                user,
+                role,
+                operator: OPERATOR_PARTY
```
dimitri-da, line 62: we would need to change this. (If in DABL we query for the .well-known parties and we use the UserAdmin as the operator party

dimitri-da (approved):
nice start! We need to make sure this works properly when deployed to dabl. We would need to:
- add a slash at the end of the `httpDataUrl` (line 34 on config.ts)
- query the well-known endpoint for the operator and public party names when we create the `UserSession` contracts

Sarahb-da (approved):
Code looks good, and will be able to further test common components when I start using them. Nice work!

PR #9, opened 2020-09-09
use react router for different pages

alexmatson-da:


ui/src/components/MainScreen.tsx
```diff
@@ -11,50 +12,34 @@ import InvestorWallet from './InvestorWallet'
...
+      >
+        <InvestorWallet/>
+      </Page>
+    </Route>
+
+    <Route path='/'>
+      <Page onLogout={onLogout}>
+        <WelcomeHeader/>
+      </Page>
+    </Route>
```
Sarahb-da, line 41: There is a specific "welcome header" for each role, so does this really ever get executed? Or, is it a few url segments deeper?

Sarahb-da (approved):

dimitri-da (approved):
lgtm

PR #11, opened 2020-09-09
Issuer Role

Sarahb-da:
Adds the issuer role, who can create asset tokens


dimitri-da (approved):
lgtm! you will need to rebase this on top of the model refactoring though which means that you will have to update your imports

ui/src/components/Issuer/Issuer.tsx
```diff
@@ -0,0 +1,38 @@
...
+import IssueAsset from './IssueAsset'
+import IssuedToken from './IssuedToken'
+
+type Props = {
+    onLogout: () => void;
+}
+
+const Issuer: React.FC<Props> = ({ onLogout }) => (
+    <Switch>
+        <Route path='/role/issuer/issue-asset'>
```
alexmatson-da, line 17: I also used absolute paths at first, but found what I think is a good pattern for making them relative. Here in the issuer component, you can do 

```
const { path, url } = useRouteMatch();
```

Then, build paths off the path value like ``<Route path={`${path}/issue-asset`}>``

For the sidenav links, pass the `url` into the sidenav and then you can do the same thing - ``<Menu.Item to={`${url}/issue-asset`} ... >`` for example

And you can use just `${url}` with nothing added, which will appropriately resolve to `/role/issuer`

alexmatson-da (commented):
Great work! One suggestion though

ui/src/components/Issuer/Issuer.tsx
Sarahb-da: ur right I will make that change - much better

alexmatson-da (approved):

PR #14, opened 2020-09-10
Add description

Sarahb-da:


daml/Marketplace/Issuer.daml
```diff
@@ -44,9 +44,10 @@ template Issuer
         with
           name : Text
           quantityPrecision : Int
+          description : Text
         do
           let tokenId = Id with signatories = fromList [issuer], label = name, version = 0
-          create Token with id = tokenId, observers = fromList [public], ..
+          create Token with id = tokenId, observers = fromList [public], description = description, ..
```
alexmatson-da, line 50: you can actually leave this line unchanged!

If the parameter name and the variable you are passing to it are the same (like you have with `description = description`), then that double dot `..` operator is some syntactic sugar that will automatically map the value to the parameter for you

alexmatson-da (approved):
Nice!

daml/Marketplace/Issuer.daml
Sarahb-da: lovely!

PR #15, opened 2020-09-10
adding an isPublic flag to tokens to control their observers

Sarahb-da:
Added `isPublic` to the token template which controls whether or not the issuer bot will automatically add observers to it.

### DIFF
--- a/automation/issuer/bot/issuer_bot.py
+++ b/automation/issuer/bot/issuer_bot.py
@@ -36,13 +36,19 @@ def main():
 
 
     @client.ledger_created(MARKETPLACE.RegisteredCustodian)
-    def handle_handle_registered_custodian(event):
+    def handle_registered_custodian(event):
         logging.info(f"{MARKETPLACE.RegisteredCustodian} created!")
         tokens = client.find_active(MARKETPLACE.Token)
         commands = []
+
         for token_cid, token in tokens.items():
             logging.info(f"token_cid: {token_cid}, token: {token}")
-            if client.party in token['id']['signatories']['textMap'] and token['isPublic']:
+
+            if not token['isPublic']:
+                logging.info(f"{MARKETPLACE.Token} is not public")
+                return
+
+            if client.party in token['id']['signatories']['textMap']:
                 commands.append(exercise(token_cid, 'Token_AddObservers',
                                          {'party': client.party,
                                           'newObservers': {'textMap': {event.cdata['custodian']: {}}}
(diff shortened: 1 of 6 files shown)


automation/issuer/bot/issuer_bot.py
```diff
@@ -72,7 +72,7 @@ def handle_handle_registered_investor(event):
         commands = []
         for token_cid, token in tokens.items():
             logging.info(f"token_cid: {token_cid}, token: {token}")
-            if client.party in token['id']['signatories']['textMap']:
+            if client.party in token['id']['signatories']['textMap'] and token['isPublic']:
```
dimitri-da, line 75: the decorated function names need changing. It looks like there is also a name collision `handle_handle_registered_investor` is used twice. I am addressing the naming in a PR so we may have a mini conflict there

automation/issuer/bot/issuer_bot.py
```diff
@@ -98,11 +98,12 @@ def handle_handle_registered_investor(event):
             if total_observers == 0:
                 return
 
-            logging.info(f"adding {total_observers} observer(s)")
-            client.submit_exercise(event.cid, 'Token_AddObservers',
-                                   {'party': client.party,
-                                    'newObservers': {'textMap': {**investors, **exchanges, **custodians}}
-                                   })
+            if event.cdata['isPublic']:
```
dimitri-da, line 101: would it make more sense to check this in the beginning of the callback so that you avoid doing all the unnecessary lookups?

ui/src/components/Issuer/IssueAsset.tsx
```diff
@@ -51,6 +53,20 @@ const IssueAsset = () => {
...
+                        <p>Public</p>
+                        <p><i>All parties will be aware of this token.</i></p>
+                    </>
+                :
+                    <>
+                        <p>Private</p>
+                        <p><i>Only a set of parties will be aware of this token.</i></p>
+                    </>}
+                <Radio toggle defaultChecked onClick={() => setIsPublic(!isPublic)}/>
+                {}
```
Sarahb-da, line 68: here I will add logic for manual list of observers

daml/Marketplace/Issuer.daml
```diff
@@ -45,9 +45,11 @@ template Issuer
           name : Text
           quantityPrecision : Int
           description : Text
+          isPublic : Bool
         do
           let tokenId = Id with signatories = fromList [issuer], label = name, version = 0
-          create Token with id = tokenId, observers = fromList [public], ..
+          let observers = if isPublic then fromList [public] else fromList[]
+          create Token with id = tokenId, observers = observers, ..
```
dimitri-da: consider
```haskell
create Token with id = tokenId, observers = if isPublic then fromList [public] else empty, ..
```

dimitri-da (commented):
nice. Not sure if you are still facing problems with this: 
> Running into an issue where the Public party is always an observer in the case that isPublic is false. It is also the only observer when isPublic is true - so I think im missing some bot action somewhere

we can talk about it. A thing that could be throwing you off is that this bot can read as the public party as well. (Take a look at where it gets initialized `party_groups=[public_party]`)

PR #21, opened 2020-09-15
Investor invite accept form

alexmatson-da:


ui/src/components/Investor/Investor.tsx
```diff
@@ -25,21 +34,41 @@ type Props = {
     onLogout: () => void;
 }
 
+const useWellKnown = () => {
```
Sarahb-da, line 37: can we make this hook common?

ui/src/components/Issuer/IssueAsset.tsx
```diff
@@ -57,19 +58,16 @@ const IssueAsset = () => {
...
+
+            <FormToggle
+                defaultChecked
+                className='issue-asset-form-item'
+                onLabel='Public'
+                onInfo='All parties will be aware of this token.'
+                offLabel='Private'
+                offInfo='Only a set of parties will be aware of this token.'
+                onClick={val => setIsPublic(val)}/>
+
```
Sarahb-da, line 70: ty

ui/src/components/Investor/Investor.tsx
```diff
@@ -66,4 +95,41 @@ const Investor: React.FC<Props> = ({ onLogout }) => {
...
 
+const InvestorSetup: React.FC = () => {
+    const [ isPublic, setIsPublic ] = useState(true);
+    const [ loading, setLoading ] = useState(false);
+    const [ error, setError ] = useState('');
+
+    const investor = useParty();
+    const ledger = useLedger();
+
+    const handleSetupSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
```
Sarahb-da, line 106: is this true for the issuer?

Sarahb-da (approved):
a few comments but looks good

ui/src/components/Investor/Investor.tsx
alexmatson-da: The issuer (and all the roles, actually) will have to accept an `IssuerInvitation` (or `{Role}Invitation`) contract. The only difference is the investor's invite contract is the only one that has any data that has to be supplied...

I've been working on the Exchange view so I've been thinking about this, and maybe for the other roles, we can just open a `useStreamFetchByKey` to listen for the invitation, and auto-exercise the choice in the frontend. If more data is added at this stage later, we can extract it out into a form then

dimitri-da (approved):
👍

PR #22, opened 2020-09-15
add observers to created asset

Sarahb-da:
if private, the user can add a list of observers to their asset

### DIFF
--- a/ui/src/components/Issuer/IssueAsset.tsx
+++ b/ui/src/components/Issuer/IssueAsset.tsx
@@ -50,41 +50,42 @@ const IssueAsset = () => {
             <div className='issue-asset-form-item'>
                 <p>Description</p>
                 <p><i>Describe the asset to potential investors.</i></p>
-                <Form.Input
+                <Form.TextArea
                     fluid
                     placeholder='description'
-                    value={description}
                     className='issue-asset-form-field'
                     onChange={e => setDescription(e.currentTarget.value)}
                 />
             </div>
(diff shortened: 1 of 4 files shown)


ui/src/components/Issuer/IssueAsset.tsx
```diff
@@ -70,6 +71,20 @@ const IssueAsset = () => {
                     </>}
                 <Radio toggle defaultChecked onClick={() => setIsPublic(!isPublic)}/>
             </div>
+
+            {!isPublic &&
+                <>
+                    <p>Add Observers</p>
+                    <p><i> Input each party Id seporated by a comma </i></p>
```
alexmatson-da, line 78: Typo, should be `separated`

ui/src/components/Issuer/IssueAsset.tsx
```diff
@@ -70,6 +71,20 @@ const IssueAsset = () => {
...
+                    <Form.Input
```
alexmatson-da, line 79: We should use a `Form.TextArea` since the party Ids on DABL are long, and people might like to list them on separate lines

```
ledger-party-aaa-aaaa-aaaa,
ledger-party-bbb-bbbb-bbbb,
ledger-party-ccc-cccc-cccc,
...
```

Then after you `.split(',')`, you should also attach a `.map(p => p.trim())` just in case the newline character or extra spaces sneak in there as well

alexmatson-da (approved):
Looks good! I think it can be merged after some comments

PR #23, opened 2020-09-16
Broker role

dimitri-da:
This is an updated take at the broker role. We don't really have to start adding front end for this yet but it is good to have it in the model to make sure we can support it.

I have added another scenario in `Main.daml` where I demonstrate a workflow in which Alice exchanges some of her BTC with USDT through a broker


alexmatson-da (approved):
Thanks for resurrecting this! This is looking good

alexgraham-da (approved):

PR #27, opened 2020-09-17
Exchange UI: View Markets and Add Pairs

alexmatson-da:


daml/Marketplace/Issuer.daml
```diff
@@ -49,7 +49,8 @@ template Issuer
           observers : [Party]
         do
           let tokenId = Id with signatories = fromList [issuer], label = name, version = 0
-          create Token with id = tokenId, observers = if isPublic then fromList [public] else fromList observers, ..
+          let tokenObservers = if isPublic then (public :: observers) else observers
```
Sarahb-da, line 52: if it is public doesnt that mean all are observers anyway?

ui/src/components/Exchange/MarketPairs.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    return (
+        <Page
+            sideNav={sideNav}
+            onLogout={onLogout}
+            menuTitle={<><PublicIcon/>Market Pairs</>}
+        >
+            <div className='market-pairs'>
+                <Card fluid className='market-pair-row card-table-header'>
+                    <div className='market-pair-column'>Pair</div>
```
Sarahb-da, line 34: fyi there are nice `Table`s in the semantic ui lib

daml/Marketplace/Issuer.daml
alexmatson-da: The only reason I changed this is because the original DAML didn't seem to work right when `isPublic` was false. The observers would get passed in, but the final Token contract didn't include them. 

So I refactored the line a little bit. Not sure why this works works when the original didn't, though

ui/src/components/Exchange/TokenSelect.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { Form } from 'semantic-ui-react'
+
+import { TokenInfo } from './Exchange'
+
+type Props = {
+    label: string;
+    className?: string;
+    selected?: string;
+    tokens: TokenInfo[];
```
Sarahb-da, line 10: maybe we should consolidate these interface types like we do in `Server.tsx`

ui/src/components/Exchange/TokenSelect.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { Form } from 'semantic-ui-react'
+
+import { TokenInfo } from './Exchange'
```
Sarahb-da, line 4: maybe we should consolidate these interface types like we do in `Server.tsx`

ui/src/components/Exchange/TokenSelect.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+    return (
+        <div className='token-select-container'>
+            <Form.Select
+                label={label}
+                className={className}
+                placeholder='Select'
+                value={selected}
+                options={options}
+                onChange={(_, result) => {
```
Sarahb-da, line 29: making this not-inline would add clarity.. maybe `handleTokenChange()`

Sarahb-da (approved):
very nice

ui/src/components/Exchange/TokenSelect.tsx
alexmatson-da: I consolidated these (along with the Tuple wrappers) into one file: `common/damlTypes.ts`

PR #28, opened 2020-09-18
Add role sumtype

alexgraham-da:
Creates a `MarketRole` type to be used for pattern matching on role types instead of on strings. Currently used in Custodian, Onboarding and RoleSelectScreen.

### DIFF
--- a/daml/Marketplace/Utils.daml
+++ b/daml/Marketplace/Utils.daml
@@ -7,6 +7,8 @@ import DA.Next.Set
 import DA.Optional
 import DA.Text
 
+data MarketRole = CustodianRole | IssuerRole | ExchangeRole | InvestorRole | BrokerRole
+    deriving (Show, Eq)
 
 getAccountId : Party -> Party -> [Party] -> Id
 getAccountId owner provider signatories = Id with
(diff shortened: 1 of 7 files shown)


daml/Marketplace/Utils.daml
```diff
@@ -7,6 +7,8 @@ import DA.Next.Set
 import DA.Optional
 import DA.Text
 
+data MarketRole = CustodianRole | IssuerRole | ExchangeRole | InvestorRole | BrokerRole
+    deriving (Show, Eq)
```
dimitri-da, line 11: 2 spaces please

ui/src/components/RoleSelectScreen.tsx
```diff
@@ -63,7 +58,7 @@ const RoleSelectScreen: React.FC<Props> = ({ onLogout }) => {
 
         setLoading(false);
         setRole(undefined);
-        history.push(`/role/${role.toLowerCase()}`);
+        history.push(`/role/${role.replace('Role', '').toLowerCase()}`);
```
dimitri-da, line 61: I believe it is better if you trim the suffix `Role` than just replace the word `Role` with the empty string

dimitri-da (approved):
lgtm. A couple of minor comments

alexmatson-da (approved):

ui/src/components/RoleSelectScreen.tsx
alexgraham-da: Changed to use `slice`.

PR #34, opened 2020-09-23
Broker UI

dimitri-da:
This PR is adding the broker role to the mix.

The broker should have a wallet where they see all their investor holdings as well as an orders page where they can inspect the order requests they get from their clients, the accepted requests that are working as well as any exchange orders that the broker has submitted on behalf of their customer(s)

This PR ~is still work in progress but it~ factors out the OrderCard and OpenOrderCard so that it can be used by both the Investor role and the Broker Role

### 2020-09-25
- Updated the broker daml model so that they can establish a relationship with a custodian
- Factored out wallet UI component
- Added deposit dropdown to fill broker orders from inventory


ui/src/components/common/Wallet.tsx
```diff
@@ -55,25 +59,30 @@ type FormProps = {
...
         try {
-            const key = wrapDamlTuple([operator, investor]);
+            const key = wrapDamlTuple([operator, party]);
             const args = { depositCid, provider: exchange };
-            await ledger.exerciseByKey(Investor.Investor_AllocateToProvider, key, args);
+            if (role === MarketRole.InvestorRole) {
+                await ledger.exerciseByKey(Investor.Investor_AllocateToProvider, key, args);
+            } else if (role === MarketRole.BrokerRole) {
+                await ledger.exerciseByKey(Broker.Broker_AllocateToProvider, key, args)
+            }
```
alexmatson-da, line 85: This is a nice pattern! My PR has different custom `CustodianRequestForm`s for the Exchange, Investor, and Issuer that are all the same apart from the exercised choice. 

I'll probably steal this approach :)

ui/src/components/common/Wallet.tsx
```diff
@@ -3,30 +3,34 @@ import { Button, Card, Header, Form } from 'semantic-ui-react'
...
 type Props = {
     deposits: DepositInfo[];
     exchanges: ExchangeInfo[];
+    role: MarketRole;
     sideNav: React.ReactElement;
     onLogout: () => void;
 }
 
-const InvestorWallet: React.FC<Props> = ({ deposits, exchanges, sideNav, onLogout }) => {
+const WalletC: React.FC<Props> = ({ deposits, exchanges, role, sideNav, onLogout }) => {
```
alexmatson-da, line 25: What does the `C` stand for in the component name?

alexmatson-da (approved):
LGTM

ui/src/components/common/OrderCard.tsx
```diff
@@ -0,0 +1,36 @@
...
+import './OrderCard.css'
+
+export type OrderProps = {
+    order: Order;
+}
+
+const OrderCard: React.FC<OrderProps> = ({ children, order }) => {
+    const pair = unwrapDamlTuple(order.pair)
+    const base = pair[0].label;
+    const quote = pair[1].label;
```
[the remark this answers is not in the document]
dimitri-da: Nice! will change

ui/src/components/common/Wallet.tsx
dimitri-da: There was a naming conflict if I let this as `Wallet`. So I changed it to `WalletC` (as wallet Component). I'll probably change the name to Holdings or something along those lines

PR #58, opened 2020-09-29
Merge and split deposits

Sarahb-da:
the next thing here is to add options for naming a split deposit etc.

### DIFF
--- a/ui/src/components/common/Holdings.tsx
+++ b/ui/src/components/common/Holdings.tsx
@@ -49,7 +49,7 @@ const Holdings: React.FC<Props> = ({ deposits, exchanges, role, sideNav, onLogou
                         const assetOptions = deposits.map(d => {
                             return {
                                 key: d.contractId,
-                                text: `${d.contractData.asset.id.label} - ${d}`,
+                                text: `${d.contractData.asset.id.label} ${d.contractData.asset.quantity} | Provider: ${d.contractData.account.provider} `,
                                 value: d.contractId
                             }
                         }).filter(k => k.key !== deposit.contractId)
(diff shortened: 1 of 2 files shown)


ui/src/components/common/Holdings.tsx
```diff
@@ -99,36 +118,117 @@ const AllocationForm: React.FC<FormProps> = ({ asset, provider, role, depositCid
...
+        if (splitAssetDecimal <= 0 ){
+            setLoading(false);
+            return setError({
+                header: 'Invalid Split Quantity',
+                message: `The splitting quantity must be greater than 0.`
+            })
+        }
+
+        try {
+            const args = { quantities: [String(splitAssetDecimal)] };
```
dimitri-da, line 162: In an ideal world we would have to get a hold of the actual token, then fetch its `quantityPrecision` and either round or not allow the user to split in more decimals than the token suggests. Unless it is easy to add we can leave as is.

dimitri-da (commented):
looks good to me. I'll let @alexmatson-da and @alexgraham-da chime in as well.
We should remove the explicit naming of `exchange`, `exchangeOptions` and "Allocate to Exchange` in these files eventually but warrants a separate future PR

ui/src/components/common/Holdings.tsx
```diff
@@ -99,36 +118,118 @@ const AllocationForm: React.FC<FormProps> = ({ asset, provider, role, depositCid
...
-                    options={options}
+                    options={exchangeOptions}
                     onChange={handleExchangeChange}/>
                 <Button
                     primary
+                    disabled={exchange == ''}
                     content='Allocate to Exchange'
                     onClick={handleDepositAllocation}/>
             </Form.Group>
+            <Form.Group className='inline-form-group action' style={{alignItems: "center"}}>
```
alexmatson-da, line 211: I think we should avoid `style` props... can this be added in the CSS file?

alexmatson-da (approved):

PR #61, opened 2020-09-30
Refactor error handled forms

alexmatson-da:


ui/src/components/common/Holdings.tsx
```diff
@@ -73,30 +72,27 @@ type FormProps = {
...
-                await ledger.exerciseByKey(Broker.Broker_AllocateToProvider, key, args)
-            }
-            clearForm()
-        } catch (err) {
-            setError(parseError(err));
+                break;
+            case MarketRole.BrokerRole:
+                await ledger.exerciseByKey(Broker.Broker_AllocateToProvider, key, args);
+                break;
+            default:
```
dimitri-da, line 91: nice

ui/src/components/Broker/BrokerOrders.tsx
```diff
@@ -136,58 +118,42 @@ type BrokerOrderCardProps = {
...
+        text: `${deposit.contractData.asset.quantity} ${deposit.contractData.asset.id.label} | ${deposit.contractData.account.id.label}`
+    }))
 
-    function clearForm() {
-        setDepositCid('')
+    const handleAcceptBrokerOrderFill = async () => {
+        const key = wrapDamlTuple([props.cdata.broker, props.cdata.brokerOrderId])
+        const args = { depositCid }
+        await ledger.exerciseByKey(BrokerOrder.BrokerOrder_Fill, key, args);
+        setDepositCid('');
```
dimitri-da, line 144: should we use `.then(setDepositCid(''))` here and in all other places where we clear the forms? (ie clear the form only on successful submission). The reasoning is that if the user inputed something that was wrong then hit submit, they would still have the option to correct it inline.

ui/src/components/Custodian/CreateDeposit.tsx
```diff
@@ -26,37 +19,26 @@ const CreateDeposit: React.FC = () => {
...
 
-    function clearForm() {
+        const tokenId = token.contractData.id;
+        const args = { beneficiary, tokenId, depositQuantity };
+        const key = wrapDamlTuple([operator, custodian]);
+        await ledger.exerciseByKey(Custodian.Custodian_CreateDeposit, key, args);
+
         setBeneficiary('')
         setToken(undefined)
         setDepositQuantity('')
```
dimitri-da, line 37: same as above here

ui/src/components/Exchange/CreateMarket.tsx
```diff
@@ -22,44 +21,31 @@ type Props = {
...
-            await ledger.exerciseByKey(Exchange.Exchange_AddPair, key, args);
-            clearForm();
-        } catch (err) {
-            setError(parseError(err));
-        }
-        setLoading(false);
+        await ledger.exerciseByKey(Exchange.Exchange_AddPair, key, args);
+
+        setBaseToken(undefined);
+        setQuoteToken(undefined);
```
dimitri-da, line 48: here as well. (you get the idea)

dimitri-da (approved):
looks good to me. Just a suggestion on field behavior when we have an error

ui/src/components/Broker/BrokerOrders.tsx
alexmatson-da: Ah, this already happens as you describe. Everything in this function gets wrapped in a `try { } except { }`, so the execution of this expands out to 

```javascript
setLoading(true);
try {
        const key = wrapDamlTuple([props.cdata.broker, props.cdata.brokerOrderId])
        const args = { depositCid }
        await ledger.exerciseByKey(BrokerOrder.BrokerOrder_Fill, key, args);
        setDepositCid('');
} catch (err) {
        setError(parseError(err));
}
setLoading(false);
```

inside the FormErrorHandled.

So if an error is caught from the ledger API call, then it won't clear

alexgraham-da (approved):

PR #67, opened 2020-09-30
Add to dropdowns

Sarahb-da:


ui/src/components/Issuer/IssueAsset.tsx
```diff
@@ -21,15 +27,27 @@ const IssueAsset = () => {
...
+                .map(ri => ({ contractId: ri.contractId, contractData: ri.payload.custodian })),
+            useStreamQueryAsPublic(RegisteredIssuer).contracts
+                .map(ri => ({ contractId: ri.contractId, contractData: ri.payload.issuer })),
+            useStreamQueryAsPublic(RegisteredInvestor).contracts
+                .map(ri => ({ contractId: ri.contractId, contractData: ri.payload.investor })),
+            useStreamQueryAsPublic(RegisteredExchange).contracts
+                .map(ri => ({ contractId: ri.contractId, contractData: ri.payload.exchange })),
+            useStreamQueryAsPublic(RegisteredBroker).contracts
+                .map(ri => ({ contractId: ri.contractId, contractData: ri.payload.broker }))
+            ].flat()
```
Sarahb-da, line 45: idk if this is technically "legal"

Sarahb-da: but if it is, then an issuer can add observers to their asset via dropdown

alexmatson-da: Looks fine to me, you're just gathering up all the publicly listed parties on the ledger :)

ui/src/components/Issuer/IssueAsset.tsx
```diff
@@ -84,14 +114,12 @@ const IssueAsset = () => {
                 offLabel='Private'
                 offInfo='Only a set of parties will be aware of this token.'
                 onClick={val => setIsPublic(val)}/>
-            <Form.TextArea
+            <Form.Select
+                multiple
```
alexmatson-da, line 118: Have you tried using a ContractSelect with `allowAdditions`? This select won't work for a user who wants to add a private party ID as an observer.

You'll have to add and pass through the `multiple` prop to it though.

ui/src/components/common/RequestCustodianRelationship.tsx
```diff
@@ -69,11 +75,16 @@ const RequestCustodianRelationship: React.FC<Props> = ({ role }) => {
...
         >
             <Form.Group className='inline-form-group'>
-                <Form.Input
-                    label='Custodian'
-                    placeholder='Enter ID'
-                    value={custodianId}
-                    onChange={e => setCustodianId(e.currentTarget.value)}/>
+                <ContractSelect
+                    className='custodian-select-container'
+                    allowAdditions
```
alexmatson-da, line 80: You probably don't need to `allowAdditions` here, because all custodians are publicly registered so there's no need to add a private custodian ID. 

You should also know that whenever making use of `allowAdditions`, you need to use the `setAddition` prop to get the added input and use it. (This is because the existing options are keyed by contractId, but added options are just user input, so they get keyed by the value itself). 

The select inside `Exchange/InviteParticipant.tsx` has an example of an addition-enabled dropdown

ui/src/components/Issuer/IssueAsset.tsx
```diff
@@ -2,14 +2,20 @@ import React, { useState } from 'react'
...
 import { wrapDamlTuple } from '../common/damlTypes'
 import { parseError, ErrorMessage } from '../common/errorTypes'
 import FormErrorHandled from '../common/FormErrorHandled'
 import FormToggle from '../common/FormToggle'
 
+import { RegisteredCustodian } from '@daml.js/da-marketplace/lib/Marketplace/Registry'
+import { RegisteredIssuer } from '@daml.js/da-marketplace/lib/Marketplace/Registry'
+import { RegisteredInvestor } from '@daml.js/da-marketplace/lib/Marketplace/Registry'
+import { RegisteredExchange } from '@daml.js/da-marketplace/lib/Marketplace/Registry'
+import { RegisteredBroker } from '@daml.js/da-marketplace/lib/Marketplace/Registry'
```
dimitri-da, line 17: could you bundle those together?
```typescript
import {
    RegisteredCustodian,
    RegisteredIssuer,
    ...
} from '@daml.js/da-marketplace/lib/Marketplace/Registry'
```

ui/src/components/Issuer/IssueAsset.tsx
```diff
@@ -21,15 +27,27 @@ const IssueAsset = () => {
...
     const [ description, setDescription ] = useState<string>('')
     const [ isPublic, setIsPublic ] = useState<boolean>(true)
-    const [ observersList, setObserversList ] = useState<string>('');
+    const [ observers, setObservers ] = useState<string[]>([]);
     const [ error, setError ] = useState<ErrorMessage>();
     const [ loading, setLoading ] = useState(false);
 
+    const allRegisteredParties = [
+            useStreamQueryAsPublic(RegisteredCustodian).contracts
+                .map(ri => ({ contractId: ri.contractId, contractData: ri.payload.custodian })),
```
dimitri-da, line 36: super minor but I presume that `ri` was an abbreviation for `RegisteredInvestor`? perhaps we can rename the variables

ui/src/components/Issuer/IssueAsset.tsx
Sarahb-da: as for this - I just went down a rabbit hole adding `multiple` to the contract selector, and u cant actually add an addition _and_ select multiple at once. So Im thinking we hold off for now and address that with potentially another input field

alexmatson-da (approved):

ui/src/components/Issuer/IssueAsset.tsx
alexmatson-da: ah I see, no worries then. thanks for investigating that!

PR #71, opened 2020-10-01
Display profiles for roles

alexmatson-da:
Adds #66 

This adds profiles to the landing page for each role, but given that it's the same info as the invite accept there's a bit of work to add formal semantics to define "Profiles" and profile fields, that I was then able to reuse for the invite form and the landing page. 

Eventually the profile page ought to be editable after signing up, but we don't have DAML choices for that yet so the form is disabled, though the info is still displayed.


ui/src/components/common/utils.ts
```diff
@@ -0,0 +1,5 @@
+const vowels = ['a', 'e', 'i', 'o', 'u'];
+
+export const indefiniteArticle = (word: string): string => {
+    return vowels.includes(word[0].toLowerCase()) ? `an ${word}` : `a ${word}`;
```
alexgraham-da, line 4: Nice 👍

alexgraham-da, line 4: Not sure how much it matters with this small a list, but using a `Set` here (with `Set.has()`) might be faster than `Array.includes()`.

alexgraham-da (approved):
Very nice, I like the way you encapsulated everything in `common/Profile.tsx`.

PR #74, opened 2020-10-01
A few bugs in master

Sarahb-da:


ui/src/components/Broker/BrokerSideNav.tsx
```diff
@@ -27,7 +27,7 @@ const ExchangeSideNav: React.FC<Props> = ({ url }) => {
                 to={url}
                 exact
             >
-                <Header as='h3'>@{registeredBroker?.payload.name || broker}</Header>
+                <Header as='h3'>@{registeredBroker?.payload.broker || broker}</Header>
```
alexgraham-da, line 30: This was `payload.name` to show the real name instead of the party ID, was there an issue with it?

Sarahb-da: Curious, I just rebased and the compiler still does not like that on the master branch

Sarahb-da: I assume it is working for you?

alexgraham-da (approved):

PR #75, opened 2020-10-02
Fix usersession bug

### LINKED ISSUE #73: Sometimes a UserSession is created with "Operator" instead of the DABL user admin party
This happens due to a race condition in which the `useWellKnownParties()` hook does not finish fetching before the user clicks the role button that they want, and the operator party defaults to the dev value which is `Operator`

alexmatson-da:
Fixes #73 

by providing an upstream patch to `dabl-react` to return status info to the caller https://github.com/digital-asset/dabl-react/pull/7 

Then implementing some loading-checking logic in the marketplace. This also adds a convenience hook `useOperator()` for the purpose of retrieving just the userAdmin -- eventually my plan is to cache the value so it doesn't have to make a network call each time, but that'll be for a future PR.


ui/src/components/Investor/Investor.tsx
```diff
@@ -13,14 +12,14 @@ import {
 } from '@daml.js/da-marketplace/lib/Marketplace/Investor'
 import { MarketRole } from '@daml.js/da-marketplace/lib/Marketplace/Utils'
 
-import { wrapDamlTuple } from '../common/damlTypes'
+import { useOperator } from '../common/common'
+import { wrapDamlTuple, damlTupleToString } from '../common/damlTypes'
```
dimitri-da, line 16: is the `damlTupleToString` import necessary?

dimitri-da (approved):
lgtm. a minor comment. Alsio should/is any of that be upstreamed back to the `dabl-react` lib?

ui/src/components/Investor/Investor.tsx
alexmatson-da: yep, it gets used below to map public registry data for the exchanges an investor is a participant of. I think its showing up as an addition because I moved the import statements around

alexgraham-da (approved):

PR #76, opened 2020-10-05
Show custodians

alexgraham-da:
- Refactored creation of `ContractInfo`
- Refactored how registry associations are done
- Added `CustodianRelationship` template
- Shows list of `CustodianRelationship`


daml/Marketplace/Custodian.daml
```diff
@@ -12,6 +12,17 @@ import DA.Action
 import DA.Next.Set
 
 
+template CustodianRelationship
+  with
+    operator : Party
+    custodian : Party
+    party : Party
```
alexmatson-da, line 19: I wonder if we should add a field that records the `Time` in which the CustodianRelationship contract instance was created?

This would allow us to add notifications to the investor view when the relationship is accepted, by checking for `CustodianRelationship`s that were created within the past X minutes from the current time.

ui/src/components/App.tsx
```diff
@@ -43,7 +44,9 @@ const App: React.FC = () => {
               >
                 <WellKnownPartiesProvider>
                   <PublicProvider>
-                    <MainScreen onLogout={() => setCredentials(undefined)}/>
+                    <RegistryLookupProvider>
+                      <MainScreen onLogout={() => setCredentials(undefined)}/>
+                    </RegistryLookupProvider>
```
alexmatson-da, line 49: nice!!

daml/Marketplace/Custodian.daml
thealexgraham: Hmmm, we could definitely add a time field (and for other contracts as well), but I'm not sure how it would work in the UI in terms of dismissing a notification, or if the relationship is accepted later in the future. 

Adding a template for catch-all notifications that only need a dismissal (as opposed to the more specific accept/reject ones) might work, like so:
```
template Notification
  with
    sender : Party
    receiver : Party
    message : Text
  where
    signatory sender, receiver
    controller receiver can
      Notification_Dismiss : ()
        do return ()
```

Which can then be created in the various `Accept` choices, with whatever text is appropriate. There's no key there because there could be a lot of instances of non-unique contracts, but we should be able to exercise them using the `contractId`. Then you just need to have a stream query of all `Notification`s similar to how `ExchangeInviteNotifications` works.

alexgraham-da: Hmmm, we could definitely add a time field (and for other contracts as well), but I'm not sure how it would work in the UI in terms of dismissing a notification, or if the relationship is accepted later in the future.

Adding a template for catch-all notifications that only need a dismissal (as opposed to the more specific accept/reject ones) might work, like so:

```
template Notification
  with
    sender : Party
    receiver : Party
    message : Text
  where
    signatory sender, receiver
    controller receiver can
      Notification_Dismiss : ()
        do return ()
```
Which can then be created in the various Accept choices, with whatever text is appropriate. There's no key there because there could be a lot of instances of non-unique contracts, but we should be able to exercise them using the contractId. Then you just need to have a stream query of all Notifications similar to how ExchangeInviteNotifications works.

alexmatson-da: 👍 explicit notification contracts is a good idea. let's add it as an issue and solve it in a future PR, just to keep this current one scoped to the custodian work

alexmatson-da (approved):

daml/Marketplace/Custodian.daml
alexgraham-da: Sounds good, I can add the issue.

dimitri-da (approved):

PR #89, opened 2020-10-07
add quant precision

Sarahb-da:
Adds quantity precision check to splitting assets

### DIFF
--- a/ui/src/components/common/Holdings.tsx
+++ b/ui/src/components/common/Holdings.tsx
@@ -204,7 +204,7 @@ const SplitForm: React.FC<SplitFormProps> = ({ deposit }) => {
 
     const tokenQuantityPercision = Number(useStreamQuery(Token).contracts
             .find(t => t.payload.id.label == deposit.contractData.asset.id.label &&
-                       t.payload.id.version === t.payload.id.version)?.payload.quantityPrecision)
+                       t.payload.id.version === t.payload.id.version)?.payload.quantityPrecision) || 0
 
     const [ splitAssetDecimal, setSplitAssetDecimal ] = useState<number>()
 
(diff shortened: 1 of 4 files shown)


ui/src/components/common/Holdings.tsx
```diff
@@ -207,36 +213,39 @@ const SplitForm: React.FC<SplitFormProps> = ({ deposit }) => {
...
 
+    const validateSplitNumber = (event: React.SyntheticEvent, result: any) => {
+        const number = Number(result.value)
+
+        if (number >= Number(asset.quantity)) {
+            return setSplitNumberError(`The splitting quantity must be less than ${asset.quantity}`)
+        }
+
+        if (number < 0) {
+            return setSplitNumberError(`The splitting quantity must be greater than 0.`)
```
dimitri-da, line 230: here the error message does not agree with the check. I think it should be `<=0`

ui/src/components/common/Holdings.tsx
```diff
@@ -207,36 +213,39 @@ const SplitForm: React.FC<SplitFormProps> = ({ deposit }) => {
...
+        }
+
+        if (countDecimals(number) > tokenQuantityPercision) {
+            return setSplitNumberError(`The decimal percision of the splitting quantity must be equal to or less than ${tokenQuantityPercision}.`)
```
dimitri-da, line 234: typo `precision`

ui/src/components/common/Holdings.tsx
```diff
@@ -199,6 +200,11 @@ type SplitFormProps = {
 const SplitForm: React.FC<SplitFormProps> = ({ deposit }) => {
     const { asset } = deposit.contractData;
     const ledger = useLedger();
+    const [ splitNumberError, setSplitNumberError ] = useState<string>()
+
+    const tokenQuantityPercision = Number(useStreamQuery(Token).contracts
```
dimitri-da, line 205: if the websocket fails here do we crash and burn? should we have a default precision or show some error?

ui/src/components/common/utils.ts
```diff
@@ -45,3 +45,9 @@ export const depositSummary = (deposits: DepositInfo[]): string => {
         .map(([label, total]) => `${label}: ${total}`)
         .join(", ");
 }
+
+export function countDecimals(value: number) {
+    if (Math.floor(value) !== value)
+        return value.toString().split(".")[1].length || 0;
```
dimitri-da, line 51: if my split quantity is `1.2000000000000000000000000000000000` will this count for 1 decimal or 34? Not really a big deal as I can always remove the trailing zeros but just wondering

dimitri-da (commented):
found a typo and a couple of minor comments

ui/src/components/common/utils.ts
Sarahb-da: the `toString()` method strips trailing zeros, so the count would be 1

ui/src/components/common/Holdings.tsx
Sarahb-da: yep will add fo rthis

dimitri-da (approved):

Issue #80, opened 2020-10-05
Add dismissible notifications 


alexgraham-da:
Certain actions such as a `CustodianRelationshipRequest` being accepted or rejected should show a notification to the requesting party when the request is resolved. These notifications are only informational, and do not need any action other than to be dismissed.

Adding a catch-all template for these types of notifications is a potential way to deal with this, as it can be reused across any choice that requires a notification:

```
template Notification
  with
    sender : Party
    receiver : Party
    message : Text
  where
    signatory sender, receiver
    controller receiver can
      Notification_Dismiss : ()
        do return ()
```

And used like this:
```
    controller custodian can
      CustodianRelationshipRequest_Approve : ContractId CustodianRelationship
        do
           ...
          create Notification with sender = custodian, receiver = requester, notification = "Custodian Relationship Approved."
          create CustodianRelationship with party = requester, created = currentTime, ..
      CustodianRelationshipRequest_Reject : ()
        do
          create Notification with sender = custodian, receiver = requester, notification = "Custodian Relationship Rejected."
          return ()


```

This can be used in any of the Accept/Reject choices, with whatever text is appropriate. There's no key there because there could be a lot of instances of non-unique contracts, but we should be able to exercise them using the `contractId`. To show the notifications, you would have a stream query of all dismissible Notifications where the current party is a `receiver`  (similar to how ExchangeInviteNotifications works).

Issue #77, opened 2020-10-05
Add BrokerCustomerInvite notifications to Investor UI


alexmatson-da:
Similar to the ExchangeParticipantInvitation notifications that exist already

PR #92, opened 2020-10-08
Add dismissible notifications

alexgraham-da:
- Added a dismissible `Notification` type in DAML and added a UI component to show all notifications (https://github.com/digital-asset/da-marketplace/issues/80)
- Added dismissible notifications for custodian relationships, exchange participants and broker customers
- Added Broker Customer Invitation notifications (https://github.com/digital-asset/da-marketplace/issues/77)
- Added list of broker customers and customer invitation to Broker Orders page. (https://github.com/digital-asset/da-marketplace/issues/90) **Note**: this can be moved elsewhere
- Added an `AcceptRejectNotification` component

### DIFF
--- a/daml/Marketplace/ExchangeParticipant.daml
+++ b/daml/Marketplace/ExchangeParticipant.daml
@@ -18,10 +18,18 @@ template ExchangeParticipantInvitation
 
     controller exchParticipant can
       ExchangeParticipantInvitation_Accept : ContractId ExchangeParticipant
-        do create ExchangeParticipant with ..
+        do
+          create Notification with
+            sender = exchParticipant, senderRole = None
+            receiver = exchange, text = "Exchange invitation accepted."
+          create ExchangeParticipant with ..
 
       ExchangeParticipantInvitation_Reject : ()
-        do return ()
+        do
+          create Notification with
+            sender = exchParticipant, senderRole = None
+            receiver = exchange, text = "Exchange invitation rejected."
+          return ()
 
 template ExchangeParticipant
   with
--- a/daml/Marketplace/ExchangeParticipant.daml
+++ b/daml/Marketplace/ExchangeParticipant.daml
@@ -18,6 +18,8 @@ template ExchangeParticipantInvitation
 
     controller exchParticipant can
       ExchangeParticipantInvitation_Accept : ContractId ExchangeParticipant
+        with
+          role: Optional MarketRole
         do
           create Notification with
             sender = exchParticipant, senderRole = None
--- a/daml/Marketplace/ExchangeParticipant.daml
+++ b/daml/Marketplace/ExchangeParticipant.daml
@@ -1,6 +1,7 @@
 daml 1.2
 module Marketplace.ExchangeParticipant where
 
+import Marketplace.Notification
 import Marketplace.Trading
 import Marketplace.Token
 import Marketplace.Utils
--- a/daml/Marketplace/ExchangeParticipant.daml
+++ b/daml/Marketplace/ExchangeParticipant.daml
@@ -18,17 +18,15 @@ template ExchangeParticipantInvitation
 
     controller exchParticipant can
       ExchangeParticipantInvitation_Accept : ContractId ExchangeParticipant
-        with
-          role: Optional MarketRole
         do
-          create Notification with
+          create DismissibleNotification with
             sender = exchParticipant, senderRole = None
             receiver = exchange, text = "Exchange invitation accepted."
           create ExchangeParticipant with ..
 
       ExchangeParticipantInvitation_Reject : ()
         do
-          create Notification with
+          create DismissibleNotification with
             sender = exchParticipant, senderRole = None
             receiver = exchange, text = "Exchange invitation rejected."
           return ()
--- a/daml/Marketplace/Notification.daml
+++ b/daml/Marketplace/Notification.daml
@@ -1,0 +1,18 @@
+module Marketplace.Notification where
+
+import Marketplace.Utils
+
+
+template DismissibleNotification
+  with
+    sender : Party
+    senderRole : Optional MarketRole
+    receiver : Party
+    text : Text
+  where
+    signatory sender
+    observer receiver
+
+    controller receiver can
+      DismissibleNotification_Dismiss : ()
+        do return ()
--- a/daml/Marketplace/Utils.daml
+++ b/daml/Marketplace/Utils.daml
@@ -7,6 +7,20 @@ import DA.Next.Set
 import DA.Optional
 import DA.Text
 
+template Notification
+  with
+    sender : Party
+    senderRole : Optional MarketRole
+    receiver : Party
+    text : Text
+  where
+    signatory sender, receiver
+
+    controller receiver can
+      Notification_Dismiss : ()
+        do return ()
+
+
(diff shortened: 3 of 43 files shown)


daml/Marketplace/ExchangeParticipant.daml
```diff
@@ -18,10 +18,18 @@ template ExchangeParticipantInvitation
 
     controller exchParticipant can
       ExchangeParticipantInvitation_Accept : ContractId ExchangeParticipant
-        do create ExchangeParticipant with ..
+        do
+          create Notification with
+            sender = exchParticipant, senderRole = None
```
alexgraham-da, line 23: Currently the Exchange invitations are only for Investors in the UI, but it looks like that will change to also allow brokers to be invited, so I left this as `None` for now.

 Once that change happens in the UI, I'd like to consider updating `ExchangeParticipantInvitation_Accept/Reject` to have an `Optional MarketRole` field that can be filled in when selecting the party for invitation (which can have knowledge of the role, similar to how allocate to provider works now). Currently it would only have an effect on the `Notification`, but that information could also be useful elsewhere.

daml/Marketplace/ExchangeParticipant.daml
```diff
@@ -18,10 +18,18 @@ template ExchangeParticipantInvitation
 
     controller exchParticipant can
       ExchangeParticipantInvitation_Accept : ContractId ExchangeParticipant
-        do create ExchangeParticipant with ..
+        do
+          create DismissibleNotification with
+            sender = exchParticipant, senderRole = None
```
alexgraham-da, line 23: Currently the Exchange invitations are only for Investors in the UI, but it looks like that will change to also allow brokers to be invited, so I left this as `None` for now.

Once that change happens in the UI, I'd like to consider updating `ExchangeParticipantInvitation_Accept/Reject` to have an `Optional MarketRole` field that can be filled in when selecting the party for invitation (which can have knowledge of the role, similar to how allocate to provider works now). Currently it would only have an effect on the `DismissibleNotification`, but that information could also be useful elsewhere.

daml/Marketplace/Utils.daml
```diff
@@ -7,6 +7,20 @@ import DA.Next.Set
...
 import DA.Text
 
+template DismissibleNotification
+  with
+    sender : Party
+    senderRole : Optional MarketRole
+    receiver : Party
+    text : Text
+  where
+    signatory sender, receiver
```
dimitri-da, line 17: I would argue that the signatory of a notification should be the sender only.

daml/Marketplace/Utils.daml
```diff
@@ -7,6 +7,20 @@ import DA.Next.Set
 import DA.Optional
 import DA.Text
 
+template DismissibleNotification
```
dimitri-da, line 10: perhaps this is better if it lives on its own module. We may end up adding more functionality in the future or make it a separate module that we import

dimitri-da (commented):
a couple of comments re DAML. I didn't focus much on the UI but am planning to give it a spin

daml/Marketplace/Utils.daml
alexgraham-da: I think you're right, I'll change the `receiver` to be an `observer`.

daml/Marketplace/Utils.daml
alexgraham-da: Agreed, I moved it to `Marketplace.Notification`.

alexgraham-da: Unless you think `Marketplace.DismissibleNotification` would be better, I figured having it be more general would allow other types of notifications to be placed there.

daml/Marketplace/Utils.daml
alexgraham-da: I didn't think the `Operator` needed to have anything to do with `Notifications`, but I'm noticing that the majority of templates do have the `operator` have a field, and many with the `operator` as a `signatory`. Would there be a reason to add the `operator` here?

ui/src/components/Broker/InviteBrokerCustomer.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    const ledger = useLedger();
+    const exchange = useParty();
+    const operator = useOperator();
+
+    const currentInvitations = useStreamQuery(BrokerCustomerInvitation).contracts.map(makeContractInfo);
+    const brokerCustomers = useStreamQuery(BrokerCustomer).contracts.map(makeContractInfo);
+
+    const customerOptions = registeredInvestors.filter(ri =>
+        !brokerCustomers.find(ep => ep.contractData.brokerCustomer === ri.contractData.investor) &&
```
alexmatson-da, line 28: Perhaps `bc` for the `find` parameter?

ui/src/components/common/DismissibleNotifications.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { useLedger,  useParty, useStreamQuery } from '@daml/react'
+import { ContractId } from '@daml/types'
+import { MarketRole } from '@daml.js/da-marketplace/lib/Marketplace/Utils'
+import {
+    DismissibleNotification as DismissibleNotificationTemplate
+} from '@daml.js/da-marketplace/lib/Marketplace/Notification'
+import { makeContractInfo, DismissibleNotificationInfo } from './damlTypes'
+import { useRegistryLookup } from '../common/RegistryLookup'
+import NotificationComponent from '../common/Notification'
+import FormErrorHandled from '../common/FormErrorHandled'
```
alexmatson-da, line 13: This component is in `common` already, so the import paths can be simplified

alexmatson-da (approved):
lgtm! minor comments

ui/src/components/common/DismissibleNotifications.tsx
alexmatson-da: (side note: I kinda wish we could customize the webpack config because then can we do import aliases as described here: https://webpack.js.org/configuration/resolve/

The reason we can't at the moment is because we bootstrapped the UI from create-daml-app, which uses create-react-app scripts to manage all the underlying configs. We could eject https://create-react-app.dev/docs/available-scripts#npm-run-eject and drop react-scripts in favor of custom webpack/babel/etcetc configs, but I'm not sure if import aliases is a convincing enough reason to do that :P )

daml/Marketplace/Notification.daml
```diff
@@ -0,0 +1,18 @@
...
+
+template DismissibleNotification
+  with
+    sender : Party
+    senderRole : Optional MarketRole
+    receiver : Party
+    text : Text
+  where
+    signatory sender
+    observer receiver
```
dimitri-da, line 14: the `observer receiver` is redundant in this case. There is already a choice that the `receiver` party can exercise which means that DAML will automatically disclose this contract to the party. Doesn't hurt however having it there

dimitri-da (approved):
looks good

ui/src/components/common/DismissibleNotifications.tsx
alexgraham-da: Ha, it does certainly look a lot nicer with the import aliases though.

daml/Marketplace/Notification.daml
alexgraham-da: Makes sense, that's good to know.

PR #96, opened 2020-10-08
Allow users to update Registry profile info

alexmatson-da:
- Removes private fields from templates
- Adds DAML choice to each Registry template to allow updates

Tested on DABL, ran into some bugs that are already present in master, but the profile changes here seem to be in working order

### DIFF
--- a/daml/Marketplace/Registry.daml
+++ b/daml/Marketplace/Registry.daml
@@ -16,6 +16,14 @@ template RegisteredCustodian
     key (operator, custodian) : (Party, Party)
     maintainer key._2
 
+    controller custodian can
+      RegisteredCustodian_UpdateProfile : ContractId RegisteredCustodian
+        with
+          newName : Text
+          newLocation : Text
+        do
+          create this with name = newName, location = newLocation
+
     controller operator can
       RegisteredCustodian_Offboard : ()
         do return ()
@@ -27,6 +35,7 @@ template RegisteredIssuer
     public : Party
     issuer : Party
     name : Text
+    location : Text
     title : Text
     issuerID : Text
   where
@@ -35,6 +44,16 @@ template RegisteredIssuer
 
     key (operator, issuer) : (Party, Party)
     maintainer key._2
+
+    controller issuer can
+      RegisteredIssuer_UpdateProfile : ContractId RegisteredIssuer
+        with
+          newName : Text
+          newLocation : Text
+          newTitle : Text
+          newIssuerID : Text
+        do
+          create this with name = newName, location = newLocation, title = newTitle, issuerID = newIssuerID
 
     controller operator can
       RegisteredIssuer_Offboard : ()
@@ -56,6 +75,14 @@ template RegisteredInvestor
     key (operator, investor) : (Party, Party)
     maintainer key._2
 
+    controller investor can
+      RegisteredInvestor_UpdateProfile : ContractId RegisteredInvestor
+        with
+          newName : Text
+          newLocation : Text
+        do
+          create this with name = newName, location = newLocation
+
     controller operator can
       RegisteredInvestor_Offboard : ()
         do return ()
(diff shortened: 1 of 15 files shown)


ui/src/components/Issuer/Issuer.tsx
```diff
@@ -36,38 +34,46 @@ const Issuer: React.FC<Props> = ({ onLogout }) => {
     const ledger = useLedger();
 
     const registeredIssuer = useStreamQuery(RegisteredIssuer);
-    const issuerModel = useStreamQuery(IssuerModel);
-
     const allCustodianRelationships = useStreamQuery(CustodianRelationship).contracts.map(makeContractInfo);
 
     const [ profile, setProfile ] = useState<Profile>({
         'name': createField('', 'Name', 'Your full legal name', 'text'),
+        'location': createField('', 'Location', 'Your current location', 'text'),
```
Sarahb-da, line 41: i feel like this should be a drop down or something - location as in zip code? State? City? not pressing but definitely vague

daml/Marketplace/Registry.daml
```diff
@@ -36,6 +45,16 @@ template RegisteredIssuer
     key (operator, issuer) : (Party, Party)
     maintainer key._2
 
+    controller issuer can
+      RegisteredIssuer_UpdateProfile : ContractId RegisteredIssuer
+        with
+          newName : Text
+          newLocation : Text
+          newTitle : Text
+          newIssuerID : Text
```
Sarahb-da, line 54: what if this party only wants to update their name? I assume then the rest of these "new" fields will actually equal the current profile configuration

ui/src/components/Issuer/IssuedToken.tsx
```diff
@@ -31,7 +31,7 @@ const IssuedToken: React.FC<Props> = ({ sideNav, onLogout }) => {
             <PageSection border='blue' background='white'>
                 <p>{token?.payload.description}</p>
 
-                <p><b>Public: </b>{token?.payload.isPublic? "True" : "False"}</p>
+                <p><b>Public: </b>{(!!token?.payload.isPublic).toString()}</p>
```
Sarahb-da, line 34: nice

ui/src/components/Issuer/Issuer.tsx
```diff
@@ -36,38 +34,46 @@ const Issuer: React.FC<Props> = ({ onLogout }) => {
...
             const riData = registeredIssuer.contracts[0].payload;
-            const issuerContract = issuerModel.contracts[0];
-            const ssn = issuerContract ? issuerContract.payload.ssn : 'Private';
             setProfile({
                 name: { ...profile.name, value: riData.name },
+                location: { ...profile.location, value: riData.location },
                 title: { ...profile.title, value: riData.title },
-                issuerID: { ...profile.issuerID, value: riData.issuerID },
-                ssn: { ...profile.ssn, value: ssn },
+                issuerID: { ...profile.issuerID, value: riData.issuerID }
```
Sarahb-da, line 53: i dont remember but is the convention in the marketplace camel case with `somethingID` or `somethignId` (like dabl)

Sarahb-da (approved):
nice work, a few minor comments

daml/Marketplace/Registry.daml
alexmatson-da: yep, the form fills in with the existing values and if you change the name field and submit, it resubmits the rest with the same original value

ui/src/components/Issuer/Issuer.tsx
alexmatson-da: the daml used `somethingID`, so I kept the convention here

alexgraham-da: Bit confusing since `somethingId` means `ContractID` by convention, but "Issuer ID" is one of the fields in a list of potential role information from a document @dimitri-da shared with me (that seems to have gotten deleted by Slack's 2 week policy).

alexgraham-da:
Looks good!

alexgraham-da (approved):

dimitri-da (approved):
nice work

PR #100, opened 2020-10-16
Use HashRouter and persist credentials to localStorage

alexmatson-da:
This PR fixes two issues: 

1) when refreshing the page on DABL, you used to get that 404 nginx error. Fixed by using HashRouter.
2) when refreshing the page (in general), you'd fall back to the login screen because credentials are not persisted. 

This also makes sure there's some type safety around the JSON-ified credentials that are read out of localStorage. 

In theory, once the token expires after 24h, then the app should redirect the user back to the login page as if it was never in the store at all, though this code path remains to be tested

### DIFF
--- a/ui/src/components/LoginScreen.tsx
+++ b/ui/src/components/LoginScreen.tsx
@@ -15,6 +15,26 @@ function useQuery() {
   return new URLSearchParams(useLocation().search);
 }
 
+function raiseParamsToHash() {
+  const url = new URL(window.location.href);
+
+  // When DABL login redirects back to app, hoist the query into the hash route.
+  // This allows react-router's HashRouter to see and parse the supplied params
+
+  // i.e., we want to turn
+  // ledgerid.projectdabl.com/?party=party&token=token/#/
+  // into
+  // ledgerid.projectdabl.com/#/?party=party&token=token
+  if (url.search !== '' && url.hash === '#/') {
+    window.location.href = `${url.origin}${url.pathname}#/${url.search}`;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


alexgraham-da (approved):
nice! works great

ui/src/components/LoginScreen.tsx
```diff
@@ -15,6 +15,26 @@ function useQuery() {
...
+  // into
+  // ledgerid.projectdabl.com/#/?party=party&token=token
+  if (url.search !== '' && url.hash === '#/') {
+    window.location.href = `${url.origin}${url.pathname}#/${url.search}`;
+  }
+}
+
+function getTokenFromCookie(): string {
+  const tokenCookiePair = document.cookie.split('; ').find(row => row.startsWith('DABL_LEDGER_ACCESS_TOKEN')) || '';
+  return tokenCookiePair.slice(tokenCookiePair.indexOf('=') + 1);
```
dimitri-da, line 35: can you clarify here why we want the part after the `=` sign? the format something like `access_token=blahblah`?

dimitri-da (approved):
looks good!

ui/src/components/LoginScreen.tsx
alexmatson-da: Yep, I believe that's the format of the cookie. Took the code from the docs (https://docs.projectdabl.com/api/iam/#the-login-in-with-dabl-button) that indicated the token delivery via URL query is going to eventually be phased out in favor of the cookie approach, so decided to get ahead on that in the marketplace

PR #113, opened 2020-11-17
Exberry pid contract

alexgraham-da:
Add a contract called `ExberryPID` that can be created to tell the exberry adapter what PID to use for order IDs. Also includes the exberry adapter in the distributed dit file.

### DIFF
--- a/dabl-meta.yaml
+++ b/dabl-meta.yaml
@@ -1,6 +1,6 @@
 catalog:
     name: da-marketplace
-    version: 0.0.2
+    version: 0.0.2.1
     short_description: DA Marketplace
     description: A marketplace for issuing and trading digital tokens.
     release_date: 2020-10-14
--- a/dabl-meta.yaml
+++ b/dabl-meta.yaml
@@ -16,4 +16,5 @@ subdeployments:
     - da-marketplace-exchange-bot-0.0.2.tar.gz
     - da-marketplace-matching-engine-0.0.2.tar.gz
     - da-marketplace-operator-bot-0.0.2.tar.gz
+    - da-marketplace-exberry-adapter-0.0.2.1.tar.gz
     - da-marketplace-ui-0.0.2.zip
--- a/dabl-meta.yaml
+++ b/dabl-meta.yaml
@@ -1,6 +1,6 @@
 catalog:
     name: da-marketplace
-    version: 0.0.2.1
+    version: 0.0.2
     short_description: DA Marketplace
     description: A marketplace for issuing and trading digital tokens.
     release_date: 2020-10-14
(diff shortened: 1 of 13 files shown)


dabl-meta.yaml
```diff
@@ -1,6 +1,6 @@
 catalog:
     name: da-marketplace
-    version: 0.0.2
+    version: 0.0.2.1
```
dimitri-da, line 3: this breaks the semantic versioning. why not make it 0.0.3?

https://semver.org/

exberry_adapter/bot/exberry_adapter_bot.py
```diff
@@ -52,6 +53,11 @@ def main():
     def say_hello(event):
         logging.info("DA Marketplace <> Exberry adapter is ready!")
 
+    @client.ledger_created(MARKETPLACE.ExberrySID)
+    def handleExberrySID(event):
```
dimitri-da, line 57: `snake_case` please

dabl-meta.yaml
alexmatson-da, line 3: Not sure how to feel about this versioning scheme... not just because it adds a 4th component, but also it leaves the other bundled packages with mismatched versions (`0.0.2` vs `0.0.2.1` for just the adapter). 

I think I would prefer simply upping everything to `0.0.3` if a version change is needed

exberry_adapter/bot/exberry_adapter_bot.py
```diff
@@ -52,6 +53,11 @@ def main():
...
+        global SID
```
dimitri-da, line 58: - What happens if you don't initialize it? it defaults to 100?
- Should the bot archive the contract after reading the info?
- Does the contract need a key to make it unique?

Please remove the # TODO at the top of the file as after this it will be done

dimitri-da (commented):
this needs a few changes

exberry_adapter/bot/exberry_adapter_bot.py
alexgraham-da: Yes - it defaults to `100`, and then after the contract is created, it changes the `SID` to whatever the contract was. If you create another contract while it's running, it will switch to that `SID`. In the README, it does point out that you have to create the contract _after_ starting the adapter.

I think it could make sense to archive the contract - after you create it and change the "starting" `SID`, that number will be out of date as soon as you perform a single order anyway.

I don't think there's a need for the contract key, especially if it gets archived. There's no need to ever lookup the contract, and since creating the contract is more of an action ("change the current starting SID to 150, please"), it doesn't matter if the contract is unique (again assuming the contract isn't archived).

dabl-meta.yaml
alexgraham-da: Yep, I can get behind this.

alexgraham-da: Leaving as `0.0.2` for now, as I plan to PR the triggers today, will change the version then.

alexmatson-da (approved):

dimitri-da (approved):

PR #115, opened 2020-11-17
Use httpBaseURL in public party provider

alexmatson-da:


alexgraham-da:
looks good! well, the code looks like the transpiled js spaghetti it is, but the PR looks good 😄

alexgraham-da (approved):

PR #116, opened 2020-11-19
Marketplace logo

alexmatson-da:
This should add the logo to the Sample Apps entry, as well as inside the app and in the tab's favicon.


alexgraham-da (approved):
nice

alexgraham-da:
Not sure if we should bump a version, or I can just replace the release since this is just to add an icon.

Issue #42, opened 2020-09-29
useStreamQuery hooks are disconnecting unexpectedly


alexmatson-da:

dimitri-da:
@alexmatson-da we verified that this also happens to the create to `create-daml-app`. perhaps we should involve the authors of `daml.js`?

alexmatson-da:
I was only able to get it to happen with `create-daml-app` once. I tried again later and couldn't get it to happen with the template. 

I've been discussing with Gary on the forum https://discuss.daml.com/t/usestreamquery-disconnecting/1325/6 and I think I'm going to try to see if a lower `reconnectThreshold` helps alleviate the issue

alexmatson-da:
Hopefully fixed with #169 , leaving open pending further testing

PR #117, opened 2020-11-20
Bump to 1.7.0

alexmatson-da:
This replaces some APIs in the UI that became deprecated in @daml/react v1.7.0, namely 
`useStreamQuery()` -> `useStreamQueries()` and 
`useStreamFetchByKey()` -> `useStreamFetchByKeys()` 

Also adds extra logging per stream query call to help with resolving #42 

Tested everything on DABL to make sure the new API calls don't introduce any new issues and that everything is otherwise working as expected.


dimitri-da (approved):

alexgraham-da:
Very nice! Thanks for doing this.

I wonder if it's worth it to do something like this to terseify all those `() => [], [], (e) => console.log(...)` etcs.

```typescript
// in utils
function logDisconnectArgs<T extends object>(name:string): [() => Query<T>[], readonly unknown[], (e: StreamCloseEvent) => void] {
    return [() => [], [], (e: StreamCloseEvent) => {
        console.log(`Unexpected close from ${name}: `, e);
    }]
}

// in use:
const allDeposits = useStreamQueries(AssetDeposit, ...logDisconnectArgs('assetDeposit')).contracts.map(makeContractInfo);

```

Would get a bit inconsistent if you actually had to use the `queryFactory` and `queryDeps` in some places though.

alexgraham-da (approved):

PR #122, opened 2020-11-24
Make Responsive

Sarahb-da:
This is the first pass of making all pages responsive and navigable at smaller widths (anything greater than 550px). Also makes side bar sticky.

next step: implement responsive ui designs


alexgraham-da (approved):
👍 👍

PR #125, opened 2020-11-30
add marketplace.svg to make package dit zip

alexgraham-da:
`marketplace.svg` needs to be added to the `.dit` zip in the root, this adds it to the `zip` stage in the `Makefile`.


alexgraham-da:
at some point I will update the `Makefile` to use `ddit`, but this will work fine in the interim.

Sarahb-da (approved):
👍

PR #126, opened 2020-12-01
Add create instrument

alexgraham-da:
Allows Exberry integration to pick up `CreateInstrumentRequest`s when a new market is created by the exchange. Also adds minimum quantity and maximum quantity to the market.

TODO: orders in exberry will fail when the pair is outside of the minimum/maximum range, but this should also be enforced by the DAML model as well. Due to visibility of the `MarketPair` contract, this will most likely necessitate changes to the automation, so I'd prefer to wait until after the triggers are merged in.

### DIFF
--- a/exberry_adapter/pyproject.toml
+++ b/exberry_adapter/pyproject.toml
@@ -6,7 +6,7 @@ authors = ["Digital Asset"]
 
 [tool.poetry.dependencies]
 python = "^3.7"
-dazl = "^6.7.4"
+dazl = "^7.3.1"
 aiohttp = "^3.6.2"
 
 [tool.poetry.dev-dependencies]
--- a/exberry_adapter/pyproject.toml
+++ b/exberry_adapter/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "bot"
-version = "0.1.4"
+version = "0.1.5"
 description = "The DA Marketplace <> Exberry Adapter Bot"
 authors = ["Digital Asset"]
 
--- a/ui/src/components/Exchange/CreateMarket.css
+++ b/ui/src/components/Exchange/CreateMarket.css
@@ -10,6 +10,11 @@
     align-items: flex-end;
 }
 
+.create-market-quantities {
+    margin-top: 10px;
+    width: 250px;
+}
+
 .create-market-options .exchange-icon {
     height: 16px;
     width: 16px;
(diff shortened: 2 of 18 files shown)


exberry_adapter/pyproject.toml
```diff
@@ -1,12 +1,12 @@
...
 name = "bot"
-version = "0.1.4"
+version = "0.1.5"
 description = "The DA Marketplace <> Exberry Adapter Bot"
 authors = ["Digital Asset"]
 
 [tool.poetry.dependencies]
 python = "^3.7"
-dazl = "^6.7.4"
+dazl = "^7.3.1"
```
alexgraham-da, line 9: DAZL needed to be updated due to a bug with deserializing enum types

ui/src/components/Exchange/CreateMarket.css
```diff
@@ -10,6 +10,11 @@
     align-items: flex-end;
 }
 
+.create-market-quantities {
+    margin-top: 10px;
+    width: 250px;
```
Sarahb-da, line 15: just make sure when setting widths it wont interfere with responsiveness (as a general reminder but this is a small width so i doubt it will here)

Sarahb-da (approved):
lgtm

ui/src/components/Exchange/CreateMarket.css
alexgraham-da: 👍  Sounds good. For this I just tried it, it looks fine at 500px (and beyond) - which is the smallest width that works before cutting off the token pair boxes. Might be better for the responsiveness if those would go on top of each other when the window is small enough:

Screen Shot 2020-12-01 at 1 48 26 PM

Sarahb-da: agreed - `flex-wrap: wrap` should do the trick

alexgraham-da: I changed `.create-market-options` to:

```
.create-market-options {
    display: flex;
    flex-direction: row;
    align-items: flex-end;
    flex-wrap: wrap;
}
```

and it works until the icon gets squeezed in and then looks like this:
Screen Shot 2020-12-01 at 2 02 52 PM

If there's a fix for this you can tell me I'll add it to the PR, but my CSS/responsiveness knowledge isn't quite deep enough to know the answer off the top of my head...

PR #127, opened 2020-12-02
add party names to custodian dropdown

alexgraham-da:
change "token" to asset in create market page, add dropdown to custodian deposit page

### DIFF
--- a/ui/src/components/Custodian/CreateDeposit.css
+++ b/ui/src/components/Custodian/CreateDeposit.css
@@ -1,8 +1,11 @@
-
 @media screen and (max-width: 1000px) {
     .custodian-clients .ui.form > .fields.inline-form-group {
         flex-direction: column;
         align-items: flex-start;
+    }
+
+    .ui.form > .fields.inline-form-group .field .create-deposit-quantity {
+        padding-left: 5px;
     }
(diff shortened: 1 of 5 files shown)


ui/src/components/Custodian/CreateDeposit.css
```diff
@@ -4,11 +4,13 @@
         align-items: flex-start;
     }
 
-    .ui.form > .fields.inline-form-group .field .create-deposit-quantity {
-        padding-left: 5px;
-    }
 
     .custodian-clients .ui.form > .fields.inline-form-group .contract-select-container {
-        padding:  12px 0 12px .5em;
+        padding:  12px 0 12px 0.5em;
```
Sarahb-da, line 9: 👍

Sarahb-da (approved):

ui/src/components/Custodian/CreateDeposit.css
alexgraham-da: Ha, I was experimenting with getting rid of it and I guess I put it back in with the `0`...I really don't have an opinion on zeros in CSS 😄 

And the trailing whitespace removals were just by reflex...

PR #130, opened 2020-12-07
fix double spending

alexgraham-da:


daml/Marketplace/ExchangeParticipant.daml
```diff
@@ -67,48 +69,78 @@ template ExchangeParticipant
...
+          debitRequestCid <- create DepositDebitRequest with
+              owner = exchParticipant
+              ownerAccount = orderDeposit.account
+              depositCid = orderDepositCid
+
+          let order = Order with
+                status = "New", orderId = -1, pair = binOption
+                asset = orderDeposit.asset, account = orderDeposit.account, ..
+          orderRequestId <- create OrderRequest with ..
+          return (orderRequestId, debitRequestCid)
```
alexmatson-da, line 87: Could the `ExchangeParticipant_MakeOrder` choice be reused here as well?

Sarahb-da (approved):

daml/Marketplace/Transfer.daml
```diff
@@ -53,3 +53,53 @@ template DepositTransferRequest
...
+    owner : Party
+    ownerAccount : Account
+    depositCid : ContractId AssetDeposit
+  where
+    let senderSig = head $ toList ownerAccount.id.signatories
+        custodian = senderSig
+    signatory owner
+
+    controller custodian can
+      DepositDebitRequest_Approve : Asset
```
alexmatson-da, line 68: Might be good to include Reject choices for completeness

alexmatson-da (approved):
Nice!

daml/Marketplace/Transfer.daml
alexgraham-da: 👍  done

daml/Marketplace/ExchangeParticipant.daml
alexgraham-da: Indeed it can, good call. I had to leave the two asserts that were in `ExchangeParticipant_MakeOrder` since the `ExchangeParticipant_PlaceBinaryOptionOrder` performs an `AssetDeposit_Split` before exercising `ExchangeParticipant_MakeOrder` (which you wouldn't want to do if it fails the asserts), but it still DRYs it up a bit 👍

PR #132, opened 2020-12-09
Add triggers

alexgraham-da:
ports the python bots to DAML triggers (minus the exberry adapter)


Makefile
```diff
@@ -22,37 +17,30 @@ daml_build_log = $(state_dir)/daml_build.log
...
 broker_log := $(state_dir)/broker.log
 
-exchange_bot_dir := automation/exchange/bot.egg-info
 exchange_pid := $(state_dir)/exchange.pid
 exchange_log := $(state_dir)/exchange.log
 
 
 ### DAML server
-.PHONY: clean stop_daml_server stop_operator stop_issuer stop_custodian stop_broker stop_exchange stop_adapter stop_matching_engine
+.PHONY: clean stop_daml_server stop_operator stop_custodian stop_broker stop_exchange stop_adapter stop_matching_engine
```
Sarahb-da, line 43: curious - why not stop issuer?

README.md
```diff
@@ -18,11 +18,12 @@ can be issued and exchanged.
 
 # Requirements
 To run the marketplace locally or to build to be deployed on DABL, the following requirements are necessary:
-1. DAML SDK (1.6)
-2. `yarn` (1.22.x)
-3. `poetry` (1.0.x)
-4. `make` (3.x)
-5. `yq` (3.x)
+1. DAML SDK (1.7)
+2. `python` (3.8)
```
Sarahb-da, line 22: 👍

Makefile
alexgraham-da: I removed the Issuer bot entirely - it now runs in the `OperatorTrigger`, the reason being the original issuer bot acted on behalf of the Public party to find the registry contracts (which triggers cannot do). Since the `UserAdmin` is a signatory/observer to almost everything, it worked just as well to put it in the operator trigger (and saves one more step of deployment configuration in the process).

README.md
```diff
@@ -155,64 +156,60 @@ daml script --participant-config participants.json --json-api --dar .daml/dist/d
...
-- `da-marketplace-issuer-bot-0.0.X.tar.gz`
-- `da-marketplace-operator-bot-0.0.X.tar.gz`
+If you have **not** bootstrapped the data, you must manually add a role contract for the `UserAdmin` party.
 
-If you are using the included matching engine instead of the [Exberry Matching Integration](#exberry-matching-engine-optional):
+On the "Live Data" tab, select the `UserAdmin` party and click "Add Contract":
+![oc01_select_add](https://user-images.githubusercontent.com/71082197/101641394-b0c3d580-39ff-11eb-94bd-6198dce25b70.png)
 
-- `da-marketplace-matching-engine-0.0.X.tar.gz`
+Then, select the `Marketplace.Operator:Operator` template, and fill in `UserAdmin` for the `Operator` field, and `Public` for the `Public` field and click "Submit":
```
alexmatson-da, line 166: Yep, not the most ideal solution but probably the best option out of what's available!

dabl-meta.yaml
```diff
@@ -10,12 +10,7 @@ catalog:
...
-    - da-marketplace-issuer-bot-0.1.6.tar.gz
-    - da-marketplace-exchange-bot-0.1.6.tar.gz
-    - da-marketplace-matching-engine-0.1.6.tar.gz
-    - da-marketplace-operator-bot-0.1.6.tar.gz
-    - da-marketplace-exberry-adapter-0.1.6.tar.gz
-    - da-marketplace-ui-0.1.6.zip
+    - da-marketplace-model-0.1.7.dar
+    - da-marketplace-triggers-0.1.7.dar
+    - da-marketplace-exberry-adapter-0.1.7.tar.gz
+    - da-marketplace-ui-0.1.7.zip
```
alexmatson-da, line 16: This is really nice!

Sarahb-da (approved):
This looks great to me 👍  I wonder if someone with more experience with triggers would also want to take a look. Possibly @lima-da ?

Really nice work!

daml/Marketplace/Issuer.daml
```diff
@@ -55,7 +55,7 @@ template Issuer
           observers : [Party]
         do
           let tokenId = Id with signatories = fromList [issuer], label = name, version = 0
-          let tokenObservers = if isPublic then (public :: observers) else observers
+          let tokenObservers = if isPublic then (public :: operator :: observers) else (operator :: observers)
```
alexmatson-da, line 58: Why are we inserting the operator here? Previously we had taken the operator out of a lot of these contracts for privacy

alexgraham-da:
> This looks great to me 👍 I wonder if someone with more experience with triggers would also want to take a look. Possibly @lima-da ?
> 
> Really nice work!

Thanks! And good point, I'll add him.

daml/Marketplace/Issuer.daml
alexgraham-da: Inserting the operator here when `isPublic` is necessary since I had to change to use the operator trigger instead of an issuer trigger (since the original issuer automation used the public party that the trigger doesn't have access to). 

So in the case where we were already adding `Public`, it's no less private...however when it isn't `isPublic`, I shouldn't be adding the operator - good catch, I will fix that.

alexmatson-da: Sounds good!

alexmatson-da (approved):

dabl-meta.yaml
lima-da: 👏  this makes the launching experience very clean

lima-da (approved):
🎉

PR #134, opened 2021-01-04
Investor deposit/withdraw and issuer observe

alexgraham-da:
Adds a withdraw/deposit choice to the investor role contract. This uses `Transfer.DepositDebitRequest` and `Transfer.DepositCreditRequest` which are automatically resolved by the custodian trigger. If we want to change it so the custodian has to manually approve each request, we may have to create separate requests in the Custodian module, as the `Trading` requests are also used by the exchange (which will most likely need to remain auto-approved).

Also fixed a few places where the issuer was being removed from the observers on asset deposits in a few places.

### DIFF
--- a/daml/Marketplace/Investor.daml
+++ b/daml/Marketplace/Investor.daml
@@ -3,6 +3,7 @@ module Marketplace.Investor where
 import Marketplace.Custodian
 import Marketplace.Registry
 import Marketplace.Transfer
+import Marketplace.Token
 import Marketplace.Utils
 
 import DA.Finance.Asset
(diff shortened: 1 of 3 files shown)


daml/Marketplace/Transfer.daml
```diff
@@ -6,7 +6,7 @@ import DA.Finance.Asset
 import DA.Finance.Asset.Settlement
 import DA.Finance.Types
 
-import DA.List
+import DA.List hiding (delete)
```
alexmatson-da, line 9: woah, what does this do? haven't seen it before!

daml/Marketplace/Investor.daml
```diff
@@ -58,6 +59,28 @@ template Investor
            assert $ deposit.account.owner == investor
            create DepositTransferRequest with sender = investor, senderAccountId = deposit.account.id, ..
 
+      nonconsuming Investor_RequestDeposit : ContractId DepositCreditRequest
```
alexmatson-da, line 62: Do these deposit credits get auto-approved by the custodian trigger? 

I'm thinking that they _should_ be auto-approved, at least for the time being, for demo-ing usability. However eventually the custodian will need to verify that a deposit is legit somehow (I'm guessing we'll need DABL integrations with external payment/bank transfer systems?)

alexmatson-da (approved):
Looks good!

Sarahb-da (approved):
👍

daml/Marketplace/Transfer.daml
alexgraham-da: `hiding` lets you import everything in a module minus the functions/templates/types you have listed there. The reason i did it here is because `DA.Next.Set` also has a `delete` function (which is the one I wanted to use) so there would be an ambiguous occurrence  error. The other way to go about this is to use `import qualified` for one of them.

daml/Marketplace/Investor.daml
alexgraham-da: Yeah it is currently auto-approved by the custodian trigger. It uses `Transfer.DepositDebitRequest` and `Transfer.DepositCreditRequest` which are shared by the exchange ordering system (which I'm pretty sure will always need them to be auto-approved), so if we need to change it so one auto-approves but not the other, we'll probably have to create a specific withdraw/deposit request contract (probably in the custodian module). I'll have to check with Eric if for demo purposes we do want them to be auto-approved.

I'd agree that payment/bank integrations would probably be the best (and niftiest!) way of handling this in the future

PR #136, opened 2021-01-06
Order amount splits

alexmatson-da:
This starts adding some improvements to the Order Form for investors. 

The main thing is instead of selecting a deposit from a dropdown, you are free to enter any amount you wish. The UI looks up available deposits and picks the first deposit CID that contains at least that amount, and exercises the same choice as it used to. 

The choice now performs the split automatically if the amount is less than the total amount of the deposit.

I also added a read-only text input that previews the total amount of asset you'll get if the order gets accepted and filled, and some nicer error handling if the amount is invalid for some reason.

Screen Shot 2021-01-06 at 11 05 06 AM
Screen Shot 2021-01-06 at 11 06 05 AM

### DIFF
--- a/daml/Marketplace/ExchangeParticipant.daml
+++ b/daml/Marketplace/ExchangeParticipant.daml
@@ -80,31 +80,47 @@ template ExchangeParticipant
         with
           depositCid : ContractId AssetDeposit
           pair : IdPair
+          amount : Decimal
           price : Decimal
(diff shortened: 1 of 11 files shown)


alexgraham-da (approved):
looks good! some really great work here.

Sarahb-da (approved):
looks good!

PR #137, opened 2021-01-07
Address and/or silence UI build warnings

alexmatson-da:


ui/src/components/Broker/Broker.tsx
```diff
@@ -85,6 +85,7 @@ const Broker: React.FC<Props> = ({ onLogout }) => {
                 location: { ...profile.location, value: rbData.location }
             })
         }
+    // eslint-disable-next-line
```
alexmatson-da, line 88: These need a bit more thought to be properly fixed; we may have to switch from `useEffect` to `useReducer` in these cases. 

So for now I just suppressed the warning as to not clutter up the console.

Sarahb-da (approved):
yay

PR #140, opened 2021-01-07
Merge deposits automatically when placing an order

alexmatson-da:


daml/Marketplace/ExchangeParticipant.daml
```diff
@@ -78,19 +78,24 @@ template ExchangeParticipant
...
       nonconsuming ExchangeParticipant_PlaceBid : (ContractId OrderRequest, ContractId DepositDebitRequest)
         with
-          depositCid : ContractId AssetDeposit
+          depositCids : [ContractId AssetDeposit]
           pair : IdPair
           amount : Decimal
           price : Decimal
         do
-          deposit <- fetch depositCid
+          aggregate_qty <- getDepositQuantities depositCids
```
alexgraham-da, line 86: Please use camel case for DAML variable names (also line 116)

daml/Marketplace/Utils.daml
```diff
@@ -40,3 +41,12 @@ getAccountOwnerProvider: Text -> (Party, Party)
 getAccountOwnerProvider label =
   let [owner, provider] = map (fromSome . partyFromText . replace "'" "") (splitOn "@" label)
   in (owner, provider)
+
+getDepositQuantities : [ContractId AssetDeposit] -> Update (Numeric 10)
```
alexgraham-da, line 45: We should use `Decimal` here - while it's just a type synonym of `Numeric 10`, `AssetDeposit` uses `Decimal` for `quantity` so it's best to match that, in case either of them get changed in the future.

daml/Marketplace/Utils.daml
```diff
@@ -40,3 +41,12 @@ getAccountOwnerProvider: Text -> (Party, Party)
...
+getDepositQuantities depositCids = do
+  deposit_quantities <- do
```
alexgraham-da, line 47: Camel case should be used here as well, also this `do` is redundant

You could also drop the variable/initial do block altogether and do `foldl (+) 0.0 <$> forA depositCids ...etc`

alexgraham-da:
a few comments on the DAML, but otherwise this looks good!

alexgraham-da (approved):

PR #144, opened 2021-01-12
add root scss and tab viewer

Sarahb-da:


ui/src/components/App.scss
```diff
@@ -0,0 +1,11 @@
+@import '../index.scss';
+@import '../colors.scss';
```
alexmatson-da, line 2: Redundant import because `index.scss` already imports `colors.scss`

alexmatson-da (approved):
looks good!

PR #146, opened 2021-01-13
Bidirectional order form

alexmatson-da:
Screen Shot 2021-01-13 at 4 07 00 PM

### DIFF
--- a/ui/src/components/Investor/InvestorTrade.scss
+++ b/ui/src/components/Investor/InvestorTrade.scss
@@ -1,13 +1,24 @@
+@import "../../index.scss";
 @import "./OrderForm.scss";
 
 @mixin investor-trade {
     .investor-trade {
         height: 100%;
 
-        .order-forms {
-            @include order-form;
-            flex: 1;
-            margin-right: var(--spacing-m);
+        .order {
+            h3 {
+                display: flex;
+                align-items: center;
+            }
+
+            .candlestick-icon {
+                margin-right: $spacing-s;
+            }
+            .order-form {
+                @include order-form;
+                flex: 1;
+                margin-right: $spacing-m;
+            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


ui/src/components/Investor/InvestorTrade.scss
```diff
@@ -0,0 +1,26 @@
...
+            .order-form {
+                @include order-form;
+                flex: 1;
+                margin-right: $spacing-m;
+            }
+        }
+    }
+}
+
+@include investor-trade;
```
Sarahb-da, line 26: Why @include this mixin here?

ui/src/icons/candlestick.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="18" height="16" viewBox="0 0 18 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Sarahb-da, line 1: do you use this?

ui/src/components/Investor/InvestorTrade.scss
Sarahb-da: possible recursive implications nature make me nervous

Sarahb-da (commented):
a few comments

ui/src/components/Investor/InvestorTrade.scss
alexmatson-da: initially, I just wanted to include it in the same file so the mixin would be used. I'll move this to the `Investor.scss` that you created after I rebase on the new changes

ui/src/icons/candlestick.svg
alexmatson-da: Not exactly; I copied the svg code into the `CandlestickIcon` component directly, but decided to keep the original file in case we needed to reference it. we can remove it if you don't think we'll need to...

alexgraham-da (approved):
nice, looks great!

ui/src/icons/candlestick.svg
Sarahb-da: Id say you can remove it

Sarahb-da (approved):

PR #149, opened 2021-01-14
Move donut table

Sarahb-da:


ui/src/components/Issuer/IssuedToken.scss
```diff
@@ -1,83 +1,104 @@
...
+@mixin issued-token {
+    .issued-token {
+        @include donut-chart;
+
+        .position-holdings-data {
+            display: flex;
+            flex-direction: row;
+
+            .ui.table.issuer-cap-table {
+                border: 0px !important;
```
alexmatson-da, line 12: `!important` makes me sad :'( 

Eventually I'd like to try to override the built-in semantic UI styles by applying higher specificity ([this](https://uxengineer.com/css-specificity-avoid-important-css/) is a neat article about it) through classes/ids/attributes/nesting rather than important 

That'll probably be a chunk of work though, so the `important` can stay for now

ui/src/components/Issuer/IssuedToken.scss
```diff
@@ -1,83 +1,104 @@
...
+            display: flex;
+            flex-direction: row;
+            justify-content: space-between;
+            align-items: center;
+
+            .token-details {
+                display: flex;
+                flex-direction: row;
+                p {
+                    margin-left: var(--spacing-m);
```
alexmatson-da, line 48: These (and a few others) should be `$spacing-m`

ui/src/components/Issuer/IssuedToken.tsx
```diff
@@ -150,4 +152,17 @@ const IssuedToken: React.FC<Props> = ({ sideNav, onLogout }) => {
...
+
+const AllocationsChart = (props: { nettedTokenDeposits: DepositInfo[] }) => {
+    return (
+        <div className='allocations'>
+            <DonutChart data={formatNetTokenDeposits(props.nettedTokenDeposits)}/>
+        </div>
+    )
+
+    function formatNetTokenDeposits(tokens: DepositInfo[]): IDonutChartData[] {
+        return tokens.map(t => {return {title: `${t.investor}/${t.provider}`, value: t.quantity, color: getDonutChartColor(tokens.indexOf(t))}})
```
alexmatson-da, line 164: suggestion: adding some new lines in here will help make this more readable, e.g. 

```suggestion
        return tokens.map(t => {
            return {
                title: `${t.investor}/${t.provider}`, 
                value: t.quantity, 
                color: getDonutChartColor(tokens.indexOf(t))
            }
        })
```

alexmatson-da (approved):
Looks good! Approved, with some musings

PR #153, opened 2021-01-16
Filter order book for current market pair orders

alexmatson-da:
This also adds [Lodash](https://lodash.com/), a very useful JS library! 

I wanted it for its `_.isEqual()` function which performs a deep equality check between two complex objects. The built-in equality operator only compares surface values, so `[1, 2, 3] === [1,2,3]` returns false, because JS sees the two arrays as two differently instantiated objects. 

On the other hand, `_.isEqual([1, 2, 3], [1, 2, 3])` returns true!


Sarahb-da (approved):
nice!

PR #155, opened 2021-01-18
Sass rework

alexmatson-da:
Most of the diff is css that is unchanged from how it was before, but reworked into mixins


ui/src/components/App.scss
```diff
@@ -1,17 +1,63 @@
...
-    width: 100vw;
+@mixin app {
+    .app {
+        height: 100vh;
+        width: 100vw;
+
+        /* common css */
+        @include card-table;
+        @include donut-chart;
+        @include form-error-handled;
```
alexmatson-da, line 36: After some consideration, I had settled on importing and `@include`ing all the common styles centrally at the top app level, since they get applied downwards

ui/src/index.scss
```diff
@@ -3,11 +3,16 @@
 
 @import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@300;400&display=swap');
 
+@import 'overrides/overrides.scss';
 @import 'colors.scss';
 @import 'icons/Icons.scss';
 
+@import 'components/App.scss';
+
+@include overrides;
```
alexmatson-da, line 12: I stuck most (though not all) of the styles that were simply changing default semantic UI classes into `overrides`

ui/src/components/App.scss
Sarahb-da: i like this approach

Sarahb-da (approved):
this is very exciting

PR #156, opened 2021-01-19
remove allocations chart + human readable names

Sarahb-da:


ui/src/components/common/utils.ts
```diff
@@ -119,3 +119,14 @@ export function useDismissableElement<T extends HTMLElement, C extends HTMLEleme
...
+
+export type IPartyInfo = {
+    party: string;
+    label: string;
+}
+
+export function getPartyLabel(partyId: string, parties: IPartyInfo[]) {
+    const partyInfo = parties.find(p => p.party === partyId)
+    console.log(partyInfo?.label)
+    return { label: partyInfo?.party || partyId , party: partyInfo?.label.substring(partyInfo.label.lastIndexOf('|')+1) || partyId}
```
alexmatson-da, line 131: I'm a bit confused by this... what's an example of what a `IPartyInfo.label` looks like? Is it just the party name or that string that looks like `"name (party) | Broker"`? 

Also, why is the `partyInfo.party` assigned to the `label` field in the return here?

alexgraham-da (approved):
looks good to me 👍

ui/src/components/common/utils.ts
Sarahb-da: forgot to remove log, and switched the label-party. Copy paste error

alexmatson-da (approved):

ui/src/components/common/utils.ts
Sarahb-da: the label contains some extra information i.e. `| Broker` - if its there, I remove it here. We can remove it all together as well but the extra info is used in some places i believe

PR #169, opened 2021-01-22
Single websocket for queries

alexmatson-da:


ui/src/components/Exchange/ExchangeParticipants.tsx
```diff
@@ -1,33 +1,28 @@
...
-import PageSection from '../common/PageSection'
-import Page from '../common/Page'
+import { useContractQuery } from '../../websocket/queryStream'
 
 import { depositSummary } from '../common/utils'
 import StripedTable from '../common/StripedTable'
+import PageSection from '../common/PageSection'
+import Page from '../common/Page'
+
+// import InviteParticipant from './InviteParticipant'
```
Sarahb-da, line 15: comment?

ui/src/components/Investor/Investor.tsx
```diff
@@ -195,9 +187,11 @@ const Investor: React.FC<Props> = ({ onLogout }) => {
...
     </Switch>
 
-    return registeredInvestor.loading
-         ? loadingScreen
-         : registeredInvestor.contracts.length === 0 ? inviteScreen : investorScreen
+    return registeredInvestor.length === 0 ? inviteScreen : investorScreen
+
+    // return registeredInvestor.loading
+    //      ? loadingScreen
+    //      : registeredInvestor.length === 0 ? inviteScreen : investorScreen
```
Sarahb-da, line 194: comment

ui/src/websocket/queryStream.ts
```diff
@@ -0,0 +1,177 @@
...
+        ...queryStream,
+        streamErrors,
+      }))
+    }
+  }, [streamErrors, queryStream.streamErrors]);
+
+  return React.createElement(QueryStreamContext.Provider, { value: queryStream }, children);
+}
+
+// export function useContractQuery
```
Sarahb-da, line 133: commeng

Sarahb-da (approved):
looks good

ui/src/components/Exchange/ExchangeParticipants.tsx
alexmatson-da: oops, ty!

ui/src/websocket/queryStream.ts
```diff
@@ -0,0 +1,177 @@
...
+export function useContractQuery<T extends object, K = unknown, I extends string = string>(template: Template<T,K,I>, asPublic?: boolean) {
```
alexgraham-da, line 134: Would it be possible to have a second function called `useContractQueryAsPublic` that just calls this one passing `true` into `asPublic` or would that cause problems with `useEffect`?

alexgraham-da (approved):
This is really great! Just one comment/question above.

Did this end up helping/solving the dropping websockets issue Eric was having?

ui/src/websocket/queryStream.ts
alexmatson-da: yep, I think that will work, but I'm not sure I see the benefit to having separate functions if the underlying implementation manages both of the public/private views anyway

alexmatson-da:
> Did this end up helping/solving the dropping websockets issue Eric was having?

It's difficult to know for sure because of how tricky it's been to repro the problem, so only time will truly tell. I have some hopes because the reconnect logic is clearer and more pronounced. Strategically, if it doesn't pan out, I still see this work as beneficial because I'm planning to try to reuse some of it when opening websockets to exberry!

ui/src/websocket/queryStream.ts
alexgraham-da: yeah - the idea was to match `useStreamQueryAsPublic` and not have to worry about importing/passing a const around, but it's really personal preference as to which which you find cleaner, so I'm fine with either.

alexgraham-da:
> > Did this end up helping/solving the dropping websockets issue Eric was having?
> 
> It's difficult to know for sure because of how tricky it's been to repro the problem, so only time will truly tell. I have some hopes because the reconnect logic is clearer and more pronounced. Strategically, if it doesn't pan out, I still see this work as beneficial because I'm planning to try to reuse some of it when opening websockets to exberry!

nice, sounds good!

PR #177, opened 2021-01-27
Public/Private toggle for creating asset, fix pie chart, fix background color, fix top-nav border to login button

Sarahb-da:


ui/src/components/Issuer/IssuedToken.tsx
```diff
@@ -138,7 +138,7 @@ const IssuedToken: React.FC<Props> = ({ sideNav, onLogout, providers, investors
                         <StripedTable
                             headings={StripedTableHeaders}
                             rows={StripedTableRows}/>
-                        {/* <AllocationsChart nettedTokenDeposits={nettedTokenDeposits}/> */}
+                        <AllocationsChart nettedTokenDeposits={nettedTokenDeposits}/>
```
alexmatson-da, line 141: yay! also could you remove the `// eslint-disable-next-line` comment down on line 189? I put it there to avoid build warnings about the component name being unused, so it should be no longer needed

alexmatson-da (approved):

PR #186, opened 2021-02-02
Listing modifications for Exberry

brianweir-da:


daml/Marketplace/Listing.daml
```diff
@@ -6,9 +6,12 @@ type T = Listing
 
 data Status
     = Active
-    | Inactive
+    | Disabled
+    | Failure with
```
lucianojoublanc-da, line 10: Don't think that failure should be captured here.
Also I've mentioned before, I would avoid making this specific to exberry.

daml/Marketplace/Listing.daml
```diff
@@ -57,30 +61,45 @@ template Service
...
         with
           createListingRequestCid : ContractId CreateListingRequest
+          providerId : Text
         do
           CreateListingRequest{..} <- fetch createListingRequestCid
           archive createListingRequestCid
           create Listing with ..
 
-      nonconsuming DeleteListing : ()
+      nonconsuming ListingFailure : ContractId CreateListingRequest
```
lucianojoublanc-da, line 80: as above; maybe return a different contract `CreateListingFailure`

lucianojoublanc-da (commented):
Good but I would split off the failure case ... I don't think it should be captured in the `State` variable.

lucianojoublanc-da (approved):

daml/Marketplace/Listing.daml
brian-weir: You're correct, this is quite specific to Exberry. My issue is that since this instrument creation with Exberry isn't documented, I don't actually know what's captured in the message vs name field so I left that in to refactor once I test with them. Plus, we need to agree on what to capture generically during failure scenarios.

daml/Marketplace/Listing.daml
brian-weir: This can easily be done but be aware that it may make the UI part more complex... We should decide if this is a design pattern we want to follow in the code base so the code is consistent everywhere.

PR #187, opened 2021-02-03
Add ccp cleared orders

alexmatson-da:


exberry_adapter/bot/exberry_adapter_bot.py
```diff
@@ -96,9 +119,18 @@ async def handle_new_order_success(event):
...
+                'orderId': event.cdata['sid']
+            }), exercise(event.cid, 'Archive', {})]
+        else:
+            req_cid, _ = await client.find_one(MARKETPLACE.ClearedOrderRequest, {
+                'order': order
+            })
+
+            return [exercise(req_cid, 'ClearedOrderRequest_Ack', {
+                'orderId': event.cdata['sid']
+            }), exercise(event.cid, 'Archive', {})]
```
alexmatson-da, line 133: I know this is a pretty clunky pattern...

alexmatson-da: Turns out this was not only clunky, but buggy... changed the approach

daml/Marketplace/ExchangeParticipant.daml
```diff
@@ -160,3 +160,20 @@ template ExchangeParticipant
           orderRequestCid <- create OrderRequest with ..
 
           return (orderRequestCid, debitRequestCid)
+
+      nonconsuming ExchangeParticipant_MakeClearedOrder : ContractId ClearedOrderRequest
+        with
+          ccp : Party
```
alexmatson-da, line 166: Where should the CCP party come from? For now I just made it a parameter of the choice, and on the UI it's currently hardcoded to `"Ccp"` when the choice is exercised. 

If we make it a dropdown on the UI that lists the CCPs that the investor is a customer of, then different investors might enter different CCPs for the same (cleared) market pair, which seems... wrong

alexgraham-da: Hmmm that's a good point. I'll have to ask Eric about customers having multiple CCPs, but I'm guessing that each cleared market will have one CCP attached to it. 

Come to think of it, the CCP already needs to know who the exchange is in order to request `FairValue`s, so I think we'll have to add a specific CCP <-> Exchange relationship.

daml/Marketplace/ExchangeParticipant.daml
```diff
@@ -160,3 +161,58 @@ template ExchangeParticipant
...
+          price : Decimal
+        do
+          (_, baseToken) <- fetchByKey @Token pair._1
+          let qty = roundBankers baseToken.quantityPrecision amount
+
+          let order = ClearedOrder with
+                isBid = False, qty, status = "New", orderId = -1, ..
+
+          exerciseByKey @CCPCustomer (ccp, operator, exchParticipant) CCPCustomer_MakeClearedOrder
+            with pair, price, order
```
alexmatson-da, line 195: I don't know if this is "elegant" or "hacky", but I made sure that an investor placing an order in a cleared market has the authorization of the CCP by moving the `MakeClearedOrder` choice to its `CCPCustomer` contract. 

if the investor is not a customer, they won't have such a contract, and the trade placement is blocked by the key not found error!

alexgraham-da: This works, you could also use `lookupByKey` which returns an `Optional` to have a clearer abort message (and maybe keep it in the `ExchangeParticipant` function):

```
optCCPCustomer <- lookupByKey @CCPCustomer etc. etc.
assertMsg "Participant must be a customer etc etc" $ isSome optCCPCustomer
```

daml/Marketplace/Exchange.daml
```diff
@@ -132,9 +136,15 @@ template Exchange
             $ minQuantity < maxQuantity
 
           assertMsg "Pair must have different base and quote tokens" $ baseToken.id.label /= quoteToken.id.label
-          let instrumentName = baseToken.id.label <> quoteToken.id.label
+
+          let instrumentName = if clearedMarket
+              -- TODO : figure out how to pad instrument names to a length of 13, then append CLR or 000
+              then baseToken.id.label <> quoteToken.id.label <> "CLR"
```
alexmatson-da, line 142: my haskell-foo is not strong enough to write a function that takes a shorter input string and a max length, and returns a padded string with as many instances of `0` to the end of the input string to fill up to the max length...

alexgraham-da: may be a nicer way but:
```
addZeros : Int -> Text -> Text
addZeros n txt = T.implode $ T.explode txt <> replicate (n - T.length txt) "0"
```

alexmatson-da: Gonna leave this as a simple "CLR" appending for now, and I'll look into creating unique instrument IDs this week / next week

ui/src/components/Investor/Investor.tsx
```diff
@@ -103,24 +104,61 @@ const Investor: React.FC<Props> = ({ onLogout }) => {
...
+                                        const tokenPair = typeof listing[0] !== 'string' && listing[0];
+                                        const defaultCCP = typeof listing[1] === 'string' && listing[1];
+
+                                        if (!tokenPair || !defaultCCP) {
+                                            throw new Error("Expected token pair and default CCP")
+                                        }
+
+                                        const [ base, quote ] = unwrapDamlTuple(tokenPair).map(t => t.label.toLowerCase());
+
+                                        console.log("Hmmmmm: ", base, quote, defaultCCP, tokenPair)
```
alexgraham-da, line 139: Can probably take this out

ui/src/components/Exchange/MarketPairs.tsx
```diff
@@ -32,7 +32,17 @@ const MarketPairs: React.FC<Props> = ({ sideNav, onLogout }) => {
...
+    const collateralizedRows = exchangeContract?.contractData.tokenPairs.map(pair => {
+        const [ base, quote ] = unwrapDamlTuple(pair);
+        const pairLabel = <>{base.label} <ExchangeIcon/> {quote.label}</>;
+        return [pairLabel, '-', '-', '-'];
+    }) || [];
+
+    const clearedRows = exchangeContract?.contractData.clearedMarkets.map(listing => {
+        const pair = unwrapDamlTuple(listing)[0];
+        if (typeof pair === 'string') {
+            throw new Error('blah')
```
alexgraham-da, line 44: Bob Loblaw's Law Blog

alexmatson-da: I was very disappointed that I had to do this, typescript can't infer the fact that this is a cleared market so the first element of the outer tuple here will always be another tuple (the Id pair)

This error would (should) never get thrown, since it's just here to satisfy the compiler... but I will change the error message though 😆

ui/src/components/Investor/Investor.tsx
alexmatson-da: Oops yeah, thought I got all of these. Thanks for catching!

alexmatson-da: Found a few more unnecessary console.logs too, removed them

ui/src/components/common/TradeCard.tsx
```diff
@@ -17,15 +25,19 @@ const TradeCard: React.FC<TradeCardProps> = ({ children, trade }) => {
     const amount = trade.isBuy ? `+ ${trade.qty} ${base}` : `- ${trade.qty} ${base}`;
     const time = new Date(0);
     // timestamp is in nanos with microsecond precision
-    time.setUTCMilliseconds(parseInt(trade.timestamp.substring(0, trade.timestamp.length - 6)))
+    if (isClearedTrade(trade)) {
+        time.setUTCMilliseconds(timestringToInt(trade.eventTimestamp))
```
alexmatson-da, line 29: as a tidbit, the ClearedTrade also has a `trade.timestamp` field like the `SettledTradeSide`, but it was in a different format (an ISO string) rather than a UNIX timestamp, so the parsing resulted in the wrong time being displayed. 

The `eventTimestamp` was as expected, but we may wish to normalize this. And I'm also not sure why we have two different timestamp fields to begin with (this was present in DerivativeTradeSide as well I believe)

daml/Marketplace/ExchangeParticipant.daml
alexmatson-da: For some reason, `lookupByKey` wasn't working even when the contract did exist. I forgot what the error was exactly (tried it a week ago), but some detail about disclosure and contract privacy I think. 

I put some frontend validation in place, so if this situation happens we stop it before exercising, and we still avoid a random DAML LF error dump on the screen.

daml/Marketplace/Trading.daml
```diff
@@ -245,6 +245,156 @@ template BrokerOrder
...
+    controller exchange can
+      ClearedOrder_Fill : Optional (ContractId ClearedOrder)
+        with
+          fillQty : Decimal
+          fillPrice : Decimal
+          counterParty : Party -- TODO Might be unneeded params
+          counterOrderId : Int
+          timestamp : Text
+        do
+          assert $ fillQty > 0.0
```
alexgraham-da, line 292: Please change these to `assertMsg`, it can get really tricky to debug with just `assert` since they give almost zero info.

ui/src/components/common/TradeCard.tsx
alexgraham-da: In `ClearedTradeSide`, `eventTimestamp` is what we receive from Exberry, and `timestamp` is a Daml `getTime` for when the trade is novated. I believe in regular `TradeSide`, `timestamp` is the `eventTimestamp` from Exberry.

So I'd say it would be good to make this more consistent, either by having the `timestamp` field in regular trades be `eventTimestamp`, or making the `timestamp` in `ClearedTradeSide` be `timeNovated` or something to that effect. Or both.

alexgraham-da (approved):
This looks good! Just make sure you pull in `master` (since there were minor changes to `add-ccp` after PR) and change the destination to `master` before merging.

ui/src/components/common/TradeCard.tsx
alexgraham-da: It might also be worth converting the `eventTimestamp` from Exberry into a Daml `Time` when creating the contracts in the adapter.

ui/src/components/Exchange/MarketPairs.tsx
alexgraham-da: Weird that it's totally fine with you assuming it's a tuple only provided it isn't a string...what if it's literally any other type??

alexmatson-da: That's the cool thing, the compiler knows at least it can't be any other type. On the Daml side, `clearedMarkets` contains items of type `((Id, Id), Party)`

and with the typescript codegen, this translates to `Tuple2<Tuple2<Id, Id>, String>` 

Then when calling `unwrapDamlTuple(listing)`, the type is converted to `(string | Tuple2<Id,Id>)[]` - an array of items that could either be strings or tuples, including the first item (although we know in advance it maps to the tuple, this may not be true just by the type) 

So in unwrapping, we actually lose some type information that we had before! Interesting, isn't it?

alexgraham-da: Ohhhhh I see! Super interesting, thank you for the explanation

PR #189, opened 2021-02-09
Add ccp

alexgraham-da:
Add CentralCounterParty and CCPCustomer. CCPCustomer maintains margin and clearing accounts with the CCP who can perform margin calls, mark to market calculations, and novate trades.


daml/Marketplace/Derivative.daml
```diff
@@ -0,0 +1,127 @@
...
+template Derivative
+  with
+    id : Id
+    pricePrecision : Int
+    description : Text
+    prodSym : Text  -- BTCF21
+    prodCode : Text  -- BTC
+    uuid : Text
+    secType : SecType
+    mmy : Text  -- YYYYMM or YYYYMMDD or YYYMMwN
```
alexmatson-da, line 37: why not a `Date`?

alexmatson-da (approved):
Looks good!

daml/Marketplace/Derivative.daml
alexgraham-da: Dimitri added this, but I'm pretty sure it's the specified date format as opposed to the actual date

alexmatson-da: Oh, I see. We might want to make it an enum or something if that's the case, but this field also doesn't seem to be in use anywhere, so we can leave it be as well

alexgraham-da: Yeah agreed. I'll do some digging to see if it will be used and how, but we'll leave it for now.

PR #193, opened 2021-02-16
New asset class model

lucianojoublanc-da:


daml/Marketplace/Custody.daml
```diff
@@ -120,6 +138,42 @@ template Service
...
+        do
+          t <- toDateUTC <$> getTime
+          LifecycleRequest{assetDepositCid, postAssetVersion} <- fetch lifecycleRequestCid
+          assetDeposit <- fetch assetDepositCid
+          (_, description) <- fetchByKey @AssetDescription assetDeposit.asset.id
+          (_, description') <- fetchByKey @AssetDescription (assetDeposit.asset.id with version = postAssetVersion)
+          let payout: Id -> Decimal -> Date -> Update (ContractId AssetDeposit)
+              payout asset qty dt = do
+                payoutDepositCid :: _ <- exercise safekeepingDepositCid $ AssetDeposit_Split [qty]
+                exerciseByKey @AssetSettlementRule description.safekeepingAccountId $
```
lucianojoublanc-da, line 155: I think we still need to call `AssetDeposit_AddObserver` here, otherwise the settlement is missing the target signatory. The reason it works in the `Tests/BinaryOption.daml` is that we add it manually (but don't remove it) when transferring the option to Bob.

daml/Tests/BinaryOption.daml
```diff
@@ -0,0 +1,87 @@
...
+  corpActReqCid <- submit alice.customer $ exerciseCmd alice.custodyServiceCid RequestCreditAccount with
+    accountId = alice.safekeepingAccountId
+    asset = usdAsset
+  corpActDepositCid <- submit common.custodian $ exerciseCmd alice.custodyServiceCid CreditAccount with 
+    creditAccountRequestCid = corpActReqCid
+
+  -- Bob exercises the Option
+  lifecycleReqCid <- submit bob.customer $ exerciseCmd bob.custodyServiceCid RequestLifecycle with
+    assetDepositCid = optionDepositCid
+    postAssetVersion = 2 -- TODO: How do we get this in practice? From a UI?
```
lucianojoublanc-da, line 83: It's very difficult to tell which version corresponds to which branch, if you just query all the `AssetDescription` contracts. We need to figure out how the user will get this argument to pass in the choice.

daml/Marketplace/Issuance.daml
```diff
@@ -51,11 +71,47 @@ template Service
           create ReduceIssuanceRequest with ..
 
     controller provider can
+      nonconsuming Originate: [ContractId AssetDescription]
+        with
+          createOriginationCid: ContractId OriginationRequest
+        do
+          OriginationRequest{..} <- fetch createOriginationCid
+          let genAssetDescriptionVersions: C -> [C]
+              genAssetDescriptionVersions = concat . para levels
```
lucianojoublanc-da, line 80: It doesn't make sense to generate new versions for all types of nodes... e.g. `When today (Scale 200.0 (One usdID))` should result in a single lifecycle request, not three ... too onerous otherwise.

daml/Tests/BinaryOption.daml
```diff
@@ -0,0 +1,87 @@
...
+  void $ submit common.custodian $ exerciseCmd bob.custodyServiceCid $ Lifecycle with
+    safekeepingDepositCid = corpActDepositCid -- TODO: how do we get this in practice?
```
lucianojoublanc-da, line 85: Currently this is stored in the `AssetDescription`, but we're picking it up from there - should that be done inside the choice? This would mean binding the safekeeping account to the asset description. Does it make sense to do that, for the sake of simplicity? What if you want to change depository? Is that possible? Would require a new asset description.

lucianojoublanc-da:
Note to self: There are numerous warnings issued in the `daml-ctl` library which we should clean-up before merging.

lucianojoublanc-da:
Also, if we merge into `master`, we'll need to open source the dependent libraries (`contingent-claims` and `daml-ctl`).

daml/Marketplace/Custody.daml
```diff
@@ -120,6 +138,42 @@ template Service
...
+                ("Lifecycle: Underlying data '" <> label <> "' missing on " <> Map.keyToText dt)
+                (Map.lookup label fixings >>= Map.lookup dt)
+              election : Claim Observation Date Id -> Claim Observation Date Id -> Claim Observation Date Id
+              election _ _ = deserialize description'.claims -- will throw if the claims don't match
+          payoutCids <- ContingentClaims.settle payout spot election (deserialize description.claims) t
+          assetDepositCid' <- create assetDeposit with asset = (assetDeposit.asset with id = (assetDeposit.asset.id with version = postAssetVersion))
+          archive assetDepositCid
+          return $ assetDepositCid' :: payoutCids
+
+-- TODO: remove when @fixings@ param above is removed; or replace with GenMap
```
lucianojoublanc-da, line 170: As noted. As of this week we're on 1.10, so try out GenMap.

daml/Tests/BinaryOption.daml
lucianojoublanc-da: I think a workaround would be to store a `AssetDescription.parentVersion`. This would make querying for the the two child branches of an `Or` trivial via the API.

daml/Marketplace/Issuance.daml
```diff
@@ -33,6 +42,17 @@ template Service
     maintainer key._1
 
     controller client can
+      nonconsuming RequestOrigination: ContractId OriginationRequest
+        with
+          assetLabel: Text
+          assetSignatories: Set Party
+          description: Text
+          issuer: Party
+          claims: Claims
```
lucianojoublanc-da, line 51: As discussed on our call, we should add `RequestOriginationOption`, `RequestOriginationBond` etc. to wrap the creation of `Claims` which are somewhat tricky via the API.
Currently this isn't a problem as we just use it in a script, where we can use the claims combinators to build the instruments, but it will become an issue when tying up with the UI.

daml/Marketplace/AssetDescription.daml
```diff
@@ -0,0 +1,33 @@
...
+    -- ^ This is the master record for the asset, and has embedded it's unique identifier
+    description: Text
+    -- ^ Short description of the asset, mainly for display purposes
+    issuer: Party
+    -- ^ The counterparty for any @claims@
+    claims: Claims
+    -- ^ Rights and obligations of the holder (and issuer)
+    safekeepingAccountId: Id
+    -- ^ Source account used for issuing the security and any subsequent corporate actions, e.g. paying out a coupon.
+    -- terms: Text
```
brian-weir, line 22: Do you plan on enabling these commented out variables in the future ? If not, remove them.

daml/Marketplace/Custody.daml
```diff
@@ -120,6 +138,22 @@ template Service
...
+        do
+          exercise lifecycleRequestCid (Lifecycle' with investor = customer, ..)
+       
+
+-- TODO: remove when @fixings@ param above is removed; or replace with GenMap
+instance MapKey Date where
+  keyToText dt = case toGregorian dt of (y,m,d) -> unwords . intersperse "-" $ show <$> [y, fromEnum m, d]
+  keyFromText s = case mapA parseInt $ splitOn "-" s of
+                    Some [y, m, d] -> fromGregorian (y, toEnum m, d)
+                    _ -> error "invalid date input"
```
brian-weir, line 155: Add the invalid key to the error output; I know the chance of this erroring is slim since its set by the `keyToText` but it does raise an error for whatever reason, then it would be good to have the input.

daml/Marketplace/Custody.daml
```diff
@@ -224,3 +258,48 @@ template TransferDepositRequest
...
+          fixings: Map Text (Map Date Decimal) --TODO these should be looked up from a contract on-ledger, not provided by the custodian!
+      controller investor  
+        do
+          t <- toDateUTC <$> getTime
+          assetDeposit <- fetch assetDepositCid
+          (_, description) <- fetchByKey @AssetDescription assetDeposit.asset.id
+          (_, description') <- fetchByKey @AssetDescription (assetDeposit.asset.id with version = postAssetVersion)
+          let payout: Id -> Decimal -> Date -> Update (ContractId AssetDeposit)
+              payout asset qty dt = do
+                payoutDepositCid :: _ <- exercise safekeepingDepositCid $ AssetDeposit_Split [qty]
```
brian-weir, line 286: Very nice way to avoid calling `head`

daml/Marketplace/Custody.daml
```diff
@@ -224,3 +258,48 @@ template TransferDepositRequest
...
+                  AssetSettlement_RemoveController description.issuer
+                return cid
+              spot : Text -> Date -> Update Decimal
+              spot label dt = pure $ fromSomeNote
+                ("Lifecycle: Underlying data '" <> label <> "' missing on " <> Map.keyToText dt)
+                (Map.lookup label fixings >>= Map.lookup dt)
+              election : Claim Observation Date Id -> Claim Observation Date Id -> Claim Observation Date Id
+              election _ _ = deserialize description'.claims -- will throw if the claims don't match
+          payoutCids <- ContingentClaims.settle payout spot election (deserialize description.claims) t
+          assetDepositCid' <- create assetDeposit with asset = (assetDeposit.asset with id = (assetDeposit.asset.id with version = postAssetVersion))
```
brian-weir, line 303: Should you not be calling `AssetSettlement_Credit` here instead of creating an asset `on the fly` ?

daml/Marketplace/Custody.daml
brian-weir: 👍

brian-weir (approved):
Great work!

daml/Marketplace/AssetDescription.daml
lucianojoublanc-da: I'm not sure. @georg-da what's your take on this? Do you think the 'shares authorized' should be a property of the origination? It protects the investors from having the issuer simply mint more tokens and diluting holdings.

I remember thinking that there was another case where this was handy, but I've forgotten it now.

daml/Marketplace/Custody.daml
lucianojoublanc-da: I'm not sure - what's the difference in this case? Let's discuss on zoom.

lucianojoublanc-da: Same goes for archiving the old asset below - should I debit instead?

daml/Marketplace/AssetDescription.daml
GeorgSchneider: I think we can revisit that topic at a later point. There needs to be a process by which an issuer sets or increases the authorized shares. Haven't thought enough about this to understand where that would fit in

daml/Marketplace/Issuance.daml
GeorgSchneider: We should think if those predefined payoffs can live in a difference service, that then uses / calls this choice on here. I could imagine such predefined payoffs to even come in a separate package, which is also how users of the marketplace can extend the functionality.

daml/Marketplace/Issuance.daml
```diff
@@ -33,6 +45,34 @@ template Service
...
+          issuer: Party
+          safekeepingAccountId: Id
+          expiry: Date
+          strike: Decimal
+          underlyingId: Id
+          currencyId: Id
+        do
+          -- TODO: refactor into a more composable form and add to contingent-claims lib
+          let binary : Decimal -> Text -> Id -> Claim O.Observation Date Id
+              binary strike spot ccy = cond (O.pure strike O.<= O.observe spot) (one currencyId) Zero
```
GeorgSchneider, line 70: this would look neater without the `O.` qualifier

daml/Marketplace/Issuance.daml
```diff
@@ -51,11 +71,47 @@ template Service
...
+                                                           signatories = assetSignatories
+                                                           version = i
+                                         ..
+
       nonconsuming CreateIssuance : (ContractId Issuance, ContractId AssetDeposit)
         with
           createIssuanceRequestCid : ContractId CreateIssuanceRequest
         do
           CreateIssuanceRequest{..} <- fetch createIssuanceRequestCid
+          Some _ <- lookupByKey @AssetDescription assetId
```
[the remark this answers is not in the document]
GeorgSchneider: y, better to add it onto the `CreateIssuanceRequest`. I'm always conscious of the Canton implications of using contract keys for anything.

daml/Tests/BinaryOption.daml
```diff
@@ -0,0 +1,87 @@
...
+      quantity = 1000.0
+  (_issuanceCid, optionDepositCid) <- submit common.custodian $ exerciseCmd alice.issuanceServiceCid $ CreateIssuance createReqCid
+
+  -- Alice gifts the option to Bob
+  addProvider <- submit bob.customer $ exerciseByKeyCmd @AssetSettlementRule bob.tradingAccountId AssetSettlement_AddController with ctrl = alice.customer
+  xferReqCid <- submit alice.customer $ exerciseCmd alice.custodyServiceCid RequestTransferDeposit with accountId = alice.safekeepingAccountId, receiverAccountId = bob.tradingAccountId, depositCid = optionDepositCid
+  optionDepositCid <- submit common.custodian $ exerciseCmd alice.custodyServiceCid $ TransferDeposit xferReqCid
+
+  -- The option expires at maturity
+  setTime $ datetime 2020 Mar 18 16 30 00
```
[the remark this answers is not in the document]
GeorgSchneider: For now yes that's fine

GeorgSchneider (approved):

daml/Marketplace/Issuance.daml
lucianojoublanc-da: Yes, the problem is that it would clash with prelude. If you look in `ContingentClaims/FinancialClaim.daml` you'll see that you can get around this by using `{-# NoImplicitPrelude #-}` but I hesitate to put that in here.

The solution is really to have this in a different file e.g. `FinancialClaim.daml`, as these are the lowest-level combinators, that I wouldn't expect a user to access directly. They would probably be using higher level combinators like `zcb` or `european` (see also my comment in the code :grin: )

lucianojoublanc-da: btw `O.<=` is `(<=) : f a -> f a -> f Bool` rather than prelude's `(<=) a -> a -> Bool`, hence the difference.

GeorgSchneider: I think you can explicitly import Prelude with an exclusion list even. Anyway, not an issue for now

PR #195, opened 2021-02-19
Primary Distribution

brianweir-da:


daml/Marketplace/Distribution/Auction/Utils.daml
```diff
@@ -0,0 +1,77 @@
...
+import Marketplace.Distribution.Auction.Model qualified as Auction
+import Marketplace.Distribution.Bid.Model qualified as Bid
+import Marketplace.Distribution.Bid.Model (Bid(..))
+import Marketplace.Distribution.Bid.Service qualified as Bid
+import Marketplace.Trading.AllocationAccountRule qualified as Allocation
+import Marketplace.Trading.AllocationAccountRule (AllocationAccountRule(..))
+
+-- | Takes a list and splits it based off a predicate. Values which match the predicate go in the right list of the tuple
+splitList : (a -> Bool) -> [a] -> ([a], [a])
+splitList p xs = foldr (\x (y,z) -> if p x then (y, x :: z) else (x :: y, z)) ([],[]) xs
```
brianweir-da, line 14: This function could output `[Either a a]` then the use could apply `partitionEithers` to get `([a],[a])` as necessary. Also, if this function is deemed useful, it could be moved to a `CommonUtils.daml` module so it can be used elsewhere.

lucianojoublanc-da: Use `DA.List.groupBy` instead.

daml/Marketplace/Distribution/Auctioneer.daml
```diff
@@ -0,0 +1,58 @@
+module Marketplace.Distribution.Auctioneer where
```
GeorgSchneider, line 1: I'd call this the `Distributor` role, it can be used later to spawn other types of distribution.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -0,0 +1,201 @@
...
+          bidCids : [ContractId Bid.T]
+        do
+          auction@Auction{..} <- fetch auctionCid
+          assertMsg "Auction already successfully processed" (status `elem` [Auction.Open, Auction.NoValidBids])
+          bids <- forA bidCids fetch
+
+          -- Split bids on those above / below the floor price -> Sort valid bids descendingly -> Execute Auction -> Split auction result on allocated / unallocated (ie, zero allocation)
+          let
+            (validBids, invalidBids) = splitList (\bid -> bid.details.price < floorPrice) bids
+            sortedBids = sortOn (\bid -> Down bid.details) validBids
```
lucianojoublanc-da, line 123: I'd maybe make this choice a wrapper on a more general function, taking a type constraint `Ord t` and constructor `(* -> t)`, so that this can be passed in by the user instead of `Down` you have here. i.e. the user can choose what ordering they want.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -0,0 +1,201 @@
...
+
+template Service
+  with
+    operator : Party
+    provider : Party
+    customer : Party
+  where
+    signatory operator, provider, customer
+
+    nonconsuming choice OfferBidder : ContractId Bid.Offer
```
GeorgSchneider, line 30: We should see if we can move this to the `Distributor` (currently `Auctioneer`) role and have the Bidding service be independent of any given auction. Would need some privacy-preserving way to invite bidders to a specific auction. Not something to do right now, but maybe leave a comment about it

daml/Marketplace/Distribution/Bid/Model.daml
```diff
@@ -0,0 +1,45 @@
+module Marketplace.Distribution.Bid.Model where
```
GeorgSchneider, line 1: Maybe rename the directory to `Bidding` instead of `Bid` - BiddingService sounds better imo.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -0,0 +1,201 @@
...
+          -- Split bids on those above / below the floor price -> Sort valid bids descendingly -> Execute Auction -> Split auction result on allocated / unallocated (ie, zero allocation)
+          let
+            (validBids, invalidBids) = splitList (\bid -> bid.details.price < floorPrice) bids
+            sortedBids = sortOn (\bid -> Down bid.details) validBids
+            ((remaining, finalPrice), allocations) = mapAccumL dutchAuction (asset.quantity, 0.0) sortedBids
+            (allocatedBids, unallocatedBids) = splitList (\allocation -> allocation.quantity == 0.0) allocations
+
+          -- Return assets of all invalid and unallocated bids. Also update their bid status accordingly
+          forA_ invalidBids $ withdrawAllocation *> updateBidStatus Bid.Invalid
+          forA_ ((.bid) `map` unallocatedBids) $ withdrawAllocation *> updateBidStatus Bid.NoAllocation
```
lucianojoublanc-da, line 129: Not a big deal, but I'm not too keen on these side effecting (i.e. returning unit) functions.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -0,0 +1,201 @@
...
+        leadBidCid : ContractId Bid.T
+      controller submitter
+      do
+        assertMsg ("Unauthorised Submitter - " <> show submitter) (submitter `elem` [provider, customer])
+        bid <- fetch leadBidCid
+        (_, auction) <- fetchByKey @Auction (bid.provider, bid.assetId)
+        assertMsg ("Unexpected lead bidder - " <> show bid.customer) (bid.customer == auction.leadBidder)
+
+        let
+          allocationPercentage = roundBankers 2 $ bid.details.quantity / auction.asset.quantity
```
GeorgSchneider, line 71: don't think we need to round here

daml/Marketplace/Distribution/Auction/Service.daml
lucianojoublanc-da: You know I'm not a fan of adding short-lived types, but in this case I might make an exception and return the invalid and not allocated bids, altogether in a new `AuctionResult`.
I think you may be doing the UI a favour by providing easy access on these, instead of having to query for them.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -0,0 +1,201 @@
...
+    asset : Asset
+    assetDepositCid : ContractId AssetDeposit
+    quotedAssetId : Id
+    allocationAccount : Account
+    receivableAccount : Account
+    floorPrice : Decimal
+  where
+    signatory operator, provider, customer
+
+template LeadBidder
```
GeorgSchneider, line 193: I'd call this something like `PublishedBid`. The issuer might decide to publicize any of the bids, doesn't have to be the lead necessarily. I think then you can also get rid of the `leadBidder` party on the auction contract.

daml/Marketplace/Distribution/Auction/Utils.daml
```diff
@@ -0,0 +1,77 @@
...
+
+-- | Returns a bidder's allocation to their trading account
+withdrawAllocation : Bid.T -> Update (ContractId AssetDeposit)
+withdrawAllocation bid@Bid{..} = do
```
lucianojoublanc-da, line 18: is the `@` used here? remove.

daml/Marketplace/Distribution/Auction/Utils.daml
```diff
@@ -0,0 +1,77 @@
...
+        -- two asset deposits - One with the allocation, one with the remaining
+        -- one asset deposit - Occurs when processing the last bid when the auction is fully allocated
+    -- 3/ Create the settlement details for delivery of the asset to the buyer
+    issuerDeposit <- fetch issuerDepositCid
+    issuerDepositCids <- exercise issuerDepositCid AssetDeposit_Split with quantities = [allocatedQuantity]
+    let
+      seller = issuer
+      (allocatedDepositCid, remainingDepositCidOpt) = case issuerDepositCids of
+        (x :: [])      -> (x, None)
+        (x :: y :: []) -> (x, Some y)
```
lucianojoublanc-da, line 70: I saw the linter here suggesting `[x]` and [x, y]` instead.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -0,0 +1,201 @@
...
+    nonconsuming choice RemoveBidder : ()
+      with
+        submitter : Party
+        bidServiceCid : ContractId Bid.S
+      controller submitter
+      do
+        assertMsg ("Unauthorised Submitter - " <> show submitter) (submitter `elem` [provider, customer])
+        archive bidServiceCid
+
+    nonconsuming choice AddLeadBidder : ContractId LeadBidder
```
GeorgSchneider, line 59: Suggest to rename that to `PublishBid` and you pass in the list of observers it should have.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -0,0 +1,201 @@
...
+      with
+        submitter : Party
+        leadBidCid : ContractId Bid.T
+      controller submitter
```
GeorgSchneider, line 63: Improvement for later: the bidder being published should probably have to agree to its bid being published. Maybe leave a comment here for now

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -0,0 +1,201 @@
...
+    signatory operator, provider, customer
+
+    nonconsuming choice OfferBidder : ContractId Bid.Offer
+      with
+        submitter : Party
+        auctionCid : ContractId Auction.T
+        bidder : Party
+      controller submitter
+      do
+        assertMsg ("Unauthorised Submitter - " <> show submitter) (submitter `elem` [provider, customer])
```
lucianojoublanc-da, line 37: I don't quite get why this is necessary.
Is it possible to instead have controller-first `provider, customer` choice statement? Wouldn't that solve this problem?

GeorgSchneider (changes requested):

daml/Marketplace/Distribution/Auction/Utils.daml
```diff
@@ -0,0 +1,77 @@
...
+
+-- | Returns a bidder's allocation to their trading account
+withdrawAllocation : Bid.T -> Update (ContractId AssetDeposit)
+withdrawAllocation bid@Bid{..} = do
+    exerciseByKey @AllocationAccountRule allocationAccount.id Allocation.Withdraw with transferTo = tradingAccount; ..
+
+-- | Updates a bid's status
+updateBidStatus : Bid.Status -> Bid.T -> Update (ContractId Bid)
+updateBidStatus newStatus bid@Bid{..} = do
+    exerciseByKey @Bid.S (bid.assetId, bid.customer) Bid.UpdateStatus with ..
```
lucianojoublanc-da, line 24: I've seen this pattern `with ..` used a lot in the marketplace, especially for Request/Propose patterns. I'm wondering whether this is in fact an anti-pattern, the alternative being passing a nested structure. i.e. instead of

```
template Proposal with a : A, b : B, c : C

Propose : Proposal with a, b, c do create Proposal ...
```
it might be better to use
```
template Proposal with result : Result
template Result with a: A, b : B, c : C

propose: Proposal with result: Result
```
I even remember seeing a related post about this in the forum; I'll try and find it.

The downside of the `..` aproach is that you end up having to create these one-line wrapper functions, instead of calling the choices directly. I'm also not sure if the compiler is smart enough to avoid de/re-constructing the data types, leading to overhead.

(Sorry, not picking on you here particularly - it's a wider thing which I think we should consider changing CCing @GeorgSchneider )

daml/Tests/Distribution/Auction.daml
```diff
@@ -0,0 +1,252 @@
...
+  (auctionCid, siCids) <- auctionAgent `submit` exerciseCmd auctionServiceCid Auction.ProcessAuction with ..
+
+  -- Assertions
+  siCids === []
+  auction <- fromSome <$> queryContractId auctionAgent auctionCid
+  auction.status === Auction.NoValidBids
+  bids <- query @Bid.Bid auctionAgent
+  all ((\bid -> bid.status == Bid.Invalid) . snd) bids === True
+
+assertAssetDeposits : Auction.Auction -> CustomerContracts -> [AssetDeposit] -> Decimal -> Decimal -> Bid.Bid -> Script ()
```
lucianojoublanc-da, line 217: Are you checking to see if rejected bids get their deposit back? I remember that was a bug in the bond-auction repo.

daml/Marketplace/Distribution/Bid/Service.daml
```diff
@@ -0,0 +1,125 @@
...
+      --     Bid{..} <- fetchAndArchive bidCid
+      --     exerciseByKey @AllocationAccountRule allocationAccount.id Allocation.Withdraw with transferTo = tradingAccount; ..
+
+    controller provider, issuer can
+      nonconsuming UpdateStatus : ContractId Bid
+        with
+          newStatus : Bid.Status
+        do
+          (bidCid, bid) <- fetchByKey @Bid (asset.id, customer)
+          archive bidCid *> create bid with status = newStatus
```
lucianojoublanc-da, line 57: You're going a bit overboard with the monad combinators. In this case I think it's just clearer to write 
```
archive bidCid 
create bid with status = newStatus
```
on separate lines

daml/Common.daml
```diff
@@ -58,10 +63,12 @@ onboardCommon = do
...
   settlementServiceOfferCid <- submit operator do exerciseCmd operatorServiceCid Operator.OfferSettlementService with provider = exchange
   settlementServiceCid <- submit exchange do exerciseCmd settlementServiceOfferCid Settlement.Accept
+  auctioneerRoleOfferCid <- submit operator do exerciseCmd operatorServiceCid Operator.OfferAuctioneerRole with provider = auctionAgent
+  auctioneerRoleCid <- submit auctionAgent do exerciseCmd auctioneerRoleOfferCid Auctioneer.Accept
   pure CommonContracts with ..
 
-onboardCustomer : CommonContracts -> Text -> Asset -> Script CustomerContracts
-onboardCustomer CommonContracts{..} party asset = do
+onboardCustomer : CommonContracts -> Text -> Asset -> Party -> Script CustomerContracts
+onboardCustomer CommonContracts{..} party asset accountNominee = do
```
lucianojoublanc-da, line 71: You mentioned this during the call. Can you explain why we need this? I thought we'd agreed that for the time being we wouldn't have a separate party for the auction, we could use the custodian as he's authorized on all the accounts?

lucianojoublanc-da (commented):
My only real blocker is the addition of the nominee to `onBoardCustomer`; the rest is cosmetic.

A wider issue I commented on, please also weigh in @georg-da @GeorgSchneider was to do with the use of `..` as an anti-pattern. But I think that should be split off as a separate issue.

daml/Marketplace/Distribution/Auction/Utils.daml
lucianojoublanc-da: Yeah, here you go: https://discuss.daml.com/t/maximum-number-of-record-fields/2073/3

daml/Marketplace/Distribution/Auction/Service.daml
GeorgSchneider: in fact I think it should be only the provider that's able to offer bidding services to the bidders

daml/Marketplace/Distribution/Auctioneer.daml
brianweir-da: Done

daml/Marketplace/Distribution/Auction/Service.daml
brianweir-da: Done

daml/Marketplace/Distribution/Bid/Model.daml
brianweir-da: Done

daml/Marketplace/Distribution/Auction/Service.daml
brianweir-da: Done

daml/Marketplace/Distribution/Auction/Service.daml
brianweir-da: Done

daml/Marketplace/Distribution/Auction/Service.daml
brianweir-da: Done

daml/Marketplace/Distribution/Auction/Model.daml
```diff
@@ -0,0 +1,50 @@
...
+    provider : Party
+    allocation : Allocation
+    price : Decimal
+    issuer : Party
+    issuerReceivableAccount : Account
+  deriving (Eq, Show)
+
+data Status
+    = Open
+    | ParticallyAllocated with
```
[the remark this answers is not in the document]
brianweir-da: Fixed - will be part of the next commit

GeorgSchneider (approved):

daml/Tests/Distribution/Auction.daml
brianweir-da: I'm adding this test in the next PR

daml/Marketplace/Distribution/Bid/Service.daml
brian-weir: done

daml/Common.daml
brian-weir: explained in Slack

daml/Marketplace/Distribution/Auction/Utils.daml
brian-weir: I thought this version was clearer than what the linter suggested but I can change it

daml/Marketplace/Distribution/Auction/Utils.daml
brian-weir: Well spotted - I've update this.

daml/Marketplace/Distribution/Auction/Service.daml
brianweir-da: As discussed - @GeorgSchneider is to look at this as he's working on the UI.

daml/Marketplace/Distribution/Auction/Utils.daml
brianweir-da: Issue has added to the list of discuss points for the overall Marketplace project

daml/Marketplace/Distribution/Auction/Service.daml
brianweir-da: The reason for its current implementation was that I presumed either a `provider` or a `customer` should be able to call these choices. If I have a controller with `provider, customer can`, then it means the caller of the choice must have both signatures in order to exercise that choice - this pattern avoids that authorisation requirement but still allows either party to call it. 

As per @GeorgSchneider 's comment, if only the `provider` can call those choices then they can be easily refactored to reflect that.

PR #198, opened 2021-02-22
Fixes from feedback

brianweir-da:


GeorgSchneider (approved):

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -120,10 +121,10 @@ template Service
 
           -- Split bids on those above / below the floor price -> Sort valid bids descendingly -> Execute Auction -> Split auction result on allocated / unallocated (ie, zero allocation)
           let
-            (validBids, invalidBids) = splitList (\bid -> bid.details.price < floorPrice) bids
+            (invalidBids, validBids) = partitionEithers $ splitList (\bid -> bid.details.price >= floorPrice) bids
```
lucianojoublanc-da, line 124: Why doesn't `goupBy` work? I don't think you need `partitionEithers`, as the left/right are the same type `a`. Something like:
```suggestion
            [invalidBids, validBids] = DA.List.groupBy (.details.price >= floorPrice) bids
```

daml/Tests/Distribution/Auction.daml
```diff
@@ -122,7 +122,11 @@ auctionParticallyAllocatedOneBidFullyAllocated = do
   -- Assertions
   length siCids === 1
   auction <- fromSome <$> queryContractId auctionAgent auctionCid
+<<<<<<< HEAD
+  auction.status === Auction.PartiallyAllocated with finalPrice = 10.0; remaining = 500.0
+=======
   auction.status === Auction.ParticallyAllocated with finalPrice = 10.0; remaining = 500.0
+>>>>>>> 86d893c5aec78c5a869287e05b87dc237a44da21
```
lucianojoublanc-da: I think you accicdentally committed an unresolved merge ...

lucianojoublanc-da (changes requested):
You committed an unfinished merge ...

daml/Marketplace/Distribution/Auction/Service.daml
brianweir-da: I commented on the chat on Friday why `groupBy` isn't ideal here - 

Hey @Luciano, I was looking at your comment on the `splitList` function -> I tried using `groupBy` but this isn’t what I’m looking for as we don't want to compare a bid to its adjacent bid in a list (it takes the function `(a -> a -> bool)`. `groupOn` looked to be what I wanted but when all elements either pass or fail the predicate (same issue with `groupBy`) => I have no easy way to know if all the elements in the list passed or failed the predicate without having to iterating again over the result.

lucianojoublanc-da (commented):
Ok for merge. I would still remove the `splitList` as it just duplicates `DA.List.groupBy`

daml/Marketplace/Distribution/Auction/Service.daml
lucianojoublanc-da: My bad, I think `groupOn` is what I wanted
```
luciano@DA-Luciano:~/src/da-marketplace$ daml repl
daml> import DA.List
daml> groupOn (>4) [1..10]
[[1,2,3,4],[5,6,7,8,9,10]]
daml> 
```

brianweir-da: Plus, you'll need a `non-exhaustive` pattern when changing the code to that suggest

lucianojoublanc-da (approved):

PR #199, opened 2021-02-22
Add UI for new modules

GeorgSchneider:
Sorry - it's a messy PR. I suggest you focus on the Daml files only.


daml/Tests/Trading/TradingIntegrationTest.daml
```diff
@@ -1,14 +1,14 @@
 module Tests.Trading.TradingIntegrationTest where
```
lucianojoublanc-da, line 1: Mmm we should remove this file if it's going to be completely commented out. @brian-weir I recall you used this for your integration tests and it will only pass when you have the adapter up, is that right? On the other hand, I do understand that we want `daml test` to pass in the CI.

lucianojoublanc-da (commented):
@GeorgSchneider I have a concern with committing `jpg` etc. binary artifacts into the source-code.

How are they doing this in the main-branch? Are they committing their resources to the git tree?

daml/Marketplace/Custody.daml
```diff
@@ -33,11 +33,11 @@ template Service
       nonconsuming RequestOpenAccount : ContractId OpenAccountRequest
         with
           accountId : Id
-          observers : Set Party
-          ctrls : Set Party
+          observers : [Party]
+          ctrls : [Party]
```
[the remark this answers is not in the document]
lucianojoublanc-da: In another file you actually did the opposite :laughing: i.e. changed List -> set, which I do agree with. :smile:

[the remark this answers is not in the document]
GeorgSchneider: Ah, this is because Sets are a pain to work with from the UI. I thought passing a list to the choice, which is then transformed into a Set right away would be ok. What do you think? Not feeling strong about this one though. Maybe the new DA.Set is represented better in TS.

daml/Tests/Trading/TradingIntegrationTest.daml
GeorgSchneider: @brian-weir 's suggestion was to move it out of the source tree that's being compiled. I think that would be a sensible option if we can get it to work as part of `daml test`. I can revert the commenting out for now and leave it for Brian to fix.

daml.yaml
```diff
@@ -1,7 +1,7 @@
 sdk-version: 1.10.0
 name: da-marketplace
 source: daml
-init-script: Setup:setup
+init-script: Demo:issuance
```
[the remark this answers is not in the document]
GeorgSchneider: Nope, good catch. Reverted

daml/Tests/Trading/TradingIntegrationTest.daml
GeorgSchneider: I've reverted this for now

daml/Marketplace/Trading/Confirmation.daml
```diff
@@ -78,4 +78,4 @@ template Confirmation
             )
 
           create Settlement.SettlementInstruction with ..
-
+
```
brian-weir, line 81: You can avoid this easily in VS Code by setting a flag which removes trailing spaces on save.

https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut

brian-weir (approved):

daml/Marketplace/Trading/Confirmation.daml
GeorgSchneider: good to know, thanks

PR #203, opened 2021-02-23
Re-introduce the bidding service and decouple bidding from the issuer

brian-weir:


daml/Marketplace/Distribution/Bidding/Model.daml
```diff
@@ -34,55 +29,19 @@ template Bid
...
     customer : Party
+    issuer : Party
     auctionId : Text
     assetId : Id
     details : Details
+    publish : Bool
     tradingAccount : Account
     allocationAccount : Account
     depositCid : ContractId AssetDeposit
     status : Status
```
brian-weir, line 41: Can remove some of the fields here as we have introduced `AuctionObserver`

daml/Marketplace/Distribution/Bidding/Request.daml
```diff
@@ -12,14 +17,13 @@ template Request
...
   where
-    signatory operator, provider, issuer
+    signatory operator, provider, customer
 
-    let assetId = asset.id
+    key (customer, auctionId) : (Party, Text)
+    maintainer key._1
 
     controller customer can
       nonconsuming SubmitBid : ContractId Bidding.Bid
```
brian-weir, line 29: Ensuring a party only submits one `Bid` is enforced by unique key constraint - A cleaner solution around this could be to add `Optional Bid` which gets set once a bid is submitted. This contract is short lived per `Auction`

daml/Marketplace/Distribution/Bidding/Model.daml
```diff
@@ -34,55 +29,19 @@ template Bid
   with
     operator : Party
     provider : Party
-    issuer : Party
     customer : Party
+    issuer : Party
     auctionId : Text
     assetId : Id
     details : Details
+    publish : Bool
```
brian-weir, line 37: Remove this from here - this should create the `PublishedBid` contract when passed into a choice with `True`. Need to think about how to make the `PublishedBid` contracts visible to all `bidders / investors`

daml/Marketplace/Distribution/Bidding/Service.daml
```diff
@@ -0,0 +1,145 @@
...
+  where
+    signatory customer
+
+    controller customer can
+      Cancel : ()
+        do pure ()
+
+    controller provider can
+      Reject : ()
+        do pure ()
```
lucianojoublanc-da, line 138: Would be useful having an error messages or something here.

lucianojoublanc-da (approved):

GeorgSchneider (approved):

PR #214, opened 2021-02-26
Transfer unallocated asset 

brian-weir:
When an auction is only partially allocated, transfer the remaining unallocated asset back to the issuer's trading account

### DIFF
--- a/daml.yaml
+++ b/daml.yaml
@@ -1,7 +1,7 @@
 sdk-version: 1.10.0
 name: da-marketplace
 source: daml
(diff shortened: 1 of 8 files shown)


lucianojoublanc-da (approved):
:+1: Do we have a policy for adding TS libraries as dependencies, or do we care at this point? I noticed @GeorgSchneider had added D3 and now you added 'moment' whatever that is ...

GeorgSchneider (approved):

PR #216, opened 2021-02-27
Refactoring of test scripts and UI improvements

GeorgSchneider:
- Tried to make the naming overall clearer, eg. onboardCommon -> onboardProviders
- Merged custodian and agent into one bank party
- Simplified accounts to: mainAccount, exchangeLockedAccount, auctionLockedAccount
- Some improvements to the claims graph in the UI


daml/Common.daml
```diff
@@ -128,12 +121,35 @@ onboardCustomer CommonContracts{..} party asset accountNominee = do
...
+    tsla = Asset with id = tslaId; quantity = 0.0
+    nflxId = Id with signatories = singleton bank; label = "NFLX"; version = 0
+    nflx = Asset with id = nflxId; quantity = 0.0
+    dUnitId = Id with signatories = singleton bank; label = "DAH-D"; version = 0
+    dUnit = Asset with id = nflxId; quantity = 0.0
+
+  submit bank do createCmd AssetDescription with assetId = usdId; description = "United States Dollar"; issuer = bank; claims = Zero; ..
+  submit bank do createCmd AssetDescription with assetId = nflxId; description = "Netflix, Inc. Common Stock"; issuer = bank; claims = Zero; ..
+  submit bank do createCmd AssetDescription with assetId = tslaId; description = "Tesla, Inc. Common Stock"; issuer = bank; claims = Zero; ..
+  submit bank do createCmd AssetDescription with assetId = dUnitId; description = "Digital Asset Holdings D-Unit"; issuer = bank; claims = Zero; ..
```
lucianojoublanc-da, line 147: Don't you have problems here because the custodian isn't a signatory? Why not use `RequestOrigination` with `claims = Zero` instead?

daml/Common.daml
```diff
@@ -33,93 +38,81 @@ data CommonContracts = CommonContracts
...
   customer <- allocatePartyWithHint party $ PartyIdHint party
   let
-    tradingAccountId      = Id with signatories = fromList [ custodian, customer ]; label = "TradingAccount-" <> party; version = 0
-    tradingAccount        = Account with provider = custodian; owner = customer; id = tradingAccountId
-    allocationAccountId   = Id with signatories = fromList [ custodian, customer ]; label = "AllocationAccount-" <> party; version = 0
-    allocationAccount     = Account with provider = custodian; owner = customer; id = allocationAccountId
-    awayAccountId         = Id with signatories = fromList [ custodian, customer ]; label = "AwayAccount-" <> party; version = 0
-    awayAccount           = Account with provider = custodian; owner = customer; id = awayAccountId
-    safekeepingAccountId  = Id with signatories = fromList [ custodian, customer ]; label = "SafekeepingAccount-" <> party; version = 0
-    safekeepingAccount    = Account with provider = custodian; owner = customer; id = safekeepingAccountId
```
lucianojoublanc-da, line 101: You've removed these. Don't you need them?

daml/Tests/Distribution/Auction.daml
```diff
@@ -17,72 +17,111 @@ import Marketplace.Distribution.Bidding.Model qualified as Bidding
...
+submitBid : Customer -> ContractId Bidding.R -> ContractId AssetDeposit -> Decimal -> Decimal -> Script (ContractId Bidding.Bid)
+submitBid party bidRequestCid depositCid price quantity = do
+  let allowPublishing = False
+  time <- getTime
+  submit party.customer do exerciseCmd party.biddingServiceCid Bidding.SubmitBid with ..
 
 auctionAndBidsFullyAllocated : Script ()
 auctionAndBidsFullyAllocated = do
-  (parties, originalDeposits, issuer, auctionAgent, custodian, auctionCid, auctionServiceCid, bidRequests) <- setup
+  Fixture{..} <- setup
```
lucianojoublanc-da, line 100: Use explicit extraction here so e.g. `Fixture{issuer,auctionAgent}` instead of `..` to make explicit used variables (some of them don't appear to be used at all).

daml/Tests/ExchangeTrade.daml
```diff
@@ -13,27 +13,28 @@ import DA.List
 
 setup : Script ()
 setup = do
-  cr@CommonContracts{..} <- onboardCommon
+  providers@Providers{..} <- onboardProviders
+  Assets{..} <- onboardAssets providers
```
lucianojoublanc-da, line 17: Again, avoid abusing `..`, except for in succinct, local scope! It makes the code much less legible.

daml/Tests/Trading/TradingIntegrationTest.daml
```diff
@@ -1,190 +1,190 @@
 module Tests.Trading.TradingIntegrationTest where
```
lucianojoublanc-da, line 1: Again, this business of commenting out the whole file. Is there some way we can remove it?

ui2/src/Apps.tsx
```diff
@@ -55,10 +55,10 @@ export default function Apps({ history } : RouteComponentProps) {
                 <Card className={classes.root}>
                   <Box border={2} borderColor="primary.main">
                     <CardActionArea onClick={() => history.push("/apps/registry/instruments")}>
-                      <CardMedia className={classes.media} image={registryImage} title="Registry" />
+                      <CardMedia className={classes.media} image={registryImage} title="Origination" />
                       <CardContent>
-                        <Typography gutterBottom variant="h5" component="h2" className={classes.cardText}>Registry</Typography>
-                        <Typography variant="body2" color="textSecondary" component="p" className={classes.cardText}>Register new instruments</Typography>
+                        <Typography gutterBottom variant="h5" component="h2" className={classes.cardText}>Origination</Typography>
+                        <Typography variant="body2" color="textSecondary" component="p" className={classes.cardText}>Manage your originations</Typography>
```
lucianojoublanc-da, line 61: New Instrument Origination?

ui2/src/Apps.tsx
```diff
@@ -73,7 +73,7 @@ export default function Apps({ history } : RouteComponentProps) {
                       <CardMedia className={classes.media} image={issuanceImage} title="Issuance" />
                       <CardContent>
                         <Typography gutterBottom variant="h5" component="h2" className={classes.cardText}>Issuance</Typography>
-                        <Typography variant="body2" color="textSecondary" component="p" className={classes.cardText}>Manage your asset issuance</Typography>
+                        <Typography variant="body2" color="textSecondary" component="p" className={classes.cardText}>Manage your issuances</Typography>
```
lucianojoublanc-da, line 76: New Instrument Issuance?

lucianojoublanc-da (commented):
I'm not convinced that renaming `custodian` -> `bank` or `away` -> `main` (account) are helpful. You're taking two specific definitions and changing them into two vague things ...

Also, I think we need to really get into the habit of avoiding `..` . It may make the code more concise, but it's more difficult to follow when there is a lot of context e.g. in a script. I don't mind using them in a short, two-liner function; but in this case, it makes it unclear where variables originate from, and whether shadowing takes place, as you can't see the names.

daml/Tests/Trading/TradingIntegrationTest.daml
brian-weir: I have taken a look at what we should do with this file. We need to agree on if we want to do integration testing via a script like this or do we want to preform it manually once the UI is up and running. Since the integration is quite difficult to setup and run on a local machine due to timing (ie, the script can only be ran after the Exberry adapter is up and running as it only picks up new events but the adapter can only be started after the Integration templates have been loaded into the sandbox, etc.) it may be better to find solution that uses the UAT environment instead. 

If we decide to keep the script, the only options we have is to move it outside of the `daml/` directory or to put it into its own folder with its own `daml.yaml` file but the file always show as having an error as it depends on `Setup:Setup` being ran before (this needs to be ran before as the Exberry Adapter and Integration depend on certain parties being created and available in the sandbox in order to be able to start them).

TL;DR - its complicated...

brian-weir (approved):

daml/Common.daml
GeorgSchneider: I use the mainAccount for the issuance now

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -39,9 +39,9 @@ template Service
           asset : Asset
           quotedAssetId : Id
           floorPrice : Decimal
-          assetDepositCid : ContractId AssetDeposit
+          depositCid : ContractId AssetDeposit
         do
-          create CreateAuctionRequest with ..
+          create CreateAuctionRequest with request = RequestCreateAuction with ..; ..
```
lucianojoublanc-da, line 44: This is not what I had in mind ... I was thinking

```
create CreateAuctionRequest with request = Auction
```

Comment continued below ...

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -54,11 +54,11 @@ template Service
...
           createAuctionRequestCid : ContractId CreateAuctionRequest
         do
-          CreateAuctionRequest{..} <- fetchAndArchive createAuctionRequestCid
+          CreateAuctionRequest{request} <- fetchAndArchive createAuctionRequestCid
 
           -- Deposit the asset into the allocation account so they cannot be transferred/withdrawn without the signatore of the Auctioneer
-          depositCid <- exerciseByKey @AllocationAccountRule allocationAccount.id Allocation.Deposit with depositCid = assetDepositCid; ..
-          create Auction with status = Auction.Open; ..
+          depositCid <- exerciseByKey @AllocationAccountRule allocationAccount.id Allocation.Deposit with depositCid = request.depositCid
+          create Auction with status = Auction.Open; auctionId = request.auctionId; asset = request.asset, quotedAssetId = request.quotedAssetId; floorPrice = request.floorPrice; ..
```
lucianojoublanc-da, line 61: ... so that here you can write `create request` that's defined above.

This way you don't need to re-list all the arguments, but you're not extracting-and-then-creating all the fields.

Does that make sense?

lucianojoublanc-da: I see that `status` is not part of the request, and I suspect that's why you haven't embedded the `request` as I suggest above.

brian-weir: Yes, the status of the `Auction` shouldn't be provided by the client. 

Also, I don't want the template for the Auction to be defined by how its created. Ie, I don't want to embed the `CreateAuctionRequest` type from the `choice` into the `Auction` template.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -157,11 +157,7 @@ template CreateAuctionRequest
     operator : Party
     provider : Party
     customer : Party
-    auctionId : Text
-    asset : Asset
-    quotedAssetId : Id
-    floorPrice : Decimal
-    assetDepositCid : ContractId AssetDeposit
+    request : RequestCreateAuction
```
brian-weir, line 160: @lucianojoublanc-da - Should we add back the individual fields so there's no need to extract and then create the fields when they are being used to create the `Auction` ?

lucianojoublanc-da: Yeah, so following on from the above comments, in this particular case, removing the `..` didn't help too much because the `Auction` type has the extra `status` field which is external. So in this case removing the `..` hasn't helped too much. I would probably leave it as it originally was, or possibly expand the `{..}` with the field names, and then do `Auction with auctionId, asset, quotedAssetId` etc. I don't know. I think this case is not as clear cut as others where we're copying every single field ...

lucianojoublanc-da: >  I would probably leave it as it originally was, or possibly expand the {..} with the field names, and then do `Auction with auctionId, asset, quotedAssetId`
 
 This is the approach we used in spider, if I recall correctly.

daml/Marketplace/Custody.daml
```diff
@@ -87,18 +87,18 @@ template Service
         with
           openAccountRequestCid : ContractId OpenAccountRequest
         do
-          OpenAccountRequest{..} <- fetchAndArchive openAccountRequestCid
-          let
-            account = Account with provider = provider; owner = customer; id = accountId
+          OpenAccountRequest{accountId; observers; ctrls} <- fetchAndArchive openAccountRequestCid
+
+          let account = Account with provider = provider; owner = customer; id = accountId
```
lucianojoublanc-da, line 92: Shorthand for `X with y = y` is `X with y` :wink:

daml/Marketplace/Registry/Issuer.daml
```diff
@@ -26,10 +26,8 @@ template Issuer
           newTitle : Text
           newIssuerID : Text
         do
-          create this with name = newName, location = newLocation, title = newTitle, issuerID = newIssuerID
+          create this with name = newName; location = newLocation; title = newTitle; issuerID = newIssuerID
```
lucianojoublanc-da, line 29: I prefer `,`. I find it really annoying that DAML has two ways of doing this!

daml/Marketplace/Trading/Order.daml
```diff
@@ -44,12 +45,10 @@ data Status
...
     | FullyExecuted
     | Rejected with
-        errorCode : Int
-        errorMessage : Text
+        reason : Error.T
     | PendingCancellation
     | CancellationRejected with
-        errorCode : Int
-        errorMessage : Text
+        reason : Error.T
```
lucianojoublanc-da, line 51: :+1: Think I mentioned this before.

brian-weir: Yup, this is based off your feedback from before - I just hadn't gotten around to updating the `Order`

daml/Marketplace/Custody.daml
brian-weir: I've updated this, thanks. - I didn't write this statement originally, I just brought up the `account = ...` line as there's only one variable being set so there was no reason for it to be on a separate line.

daml/Marketplace/Registry/Issuer.daml
brian-weir: I used to prefer `,` but I wanted to be consistent with what @GeorgSchneider had done in the project and now I prefer `;` as I'm used to it now ...

daml/Marketplace/Custody.daml
```diff
@@ -121,21 +121,21 @@ template Service
...
+          CreditAccountRequest{accountId; asset} <- fetchAndArchive creditAccountRequestCid
           exerciseByKey @AssetSettlementRule accountId AssetSettlement_Credit with ctrl = provider; ..
 
       nonconsuming DebitAccount : Asset
         with
           debitAccountRequestCid : ContractId DebitAccountRequest
         do
-          DebitAccountRequest{..} <- fetchAndArchive debitAccountRequestCid
+          DebitAccountRequest{accountId; depositCid} <- fetchAndArchive debitAccountRequestCid
           exerciseByKey @AssetSettlementRule accountId AssetSettlement_Debit with ..
```
lucianojoublanc-da, line 132: Right, so this is a good example of what I meant before, as well as the next change on line 138 ... here we can change this to:

```
data DebitAccountRequest { accountId: Id, request : AssetSettlement_debit }

...

      nonconsuming DebitAccount : Asset
        with
          debitAccountRequestCid : ContractId DebitAccountRequest
        do
          DebitAccountRequest{accountId, request} <- fetchAndArchive debitAccountRequestCid
          exerciseByKey @AssetSettlementRule accountId request
```
or even

I don't want to dwell on this - you're spending a lot of time doing this refactoring, but I just wanted to make clear what I was suggesting.

lucianojoublanc-da (approved):
Looks good. Please take a look at the comment https://github.com/digital-asset/da-marketplace/pull/216/commits/20c63687799c982a8eda215eed9bac7d5b6c0445#r584812976

I think that's an example of what I had in mind.

For now, this looks much more readable though :+1:

daml/Marketplace/Custody.daml
brian-weir: I've made this change. Looking at other services, it's a bit tricker to know if it makes sense to use it as most choices have more logic than this simple proxy example.

PR #218, opened 2021-03-02
New Origination and Lifecycling

lucianojoublanc-da:
This has a new API for lifecycling.

The investor now passes a `Claim` rather than a version, when doing the lifecyling. This means that the versioning is entirely internal (i.e. on-ledger) and not passed over the API. Otherwise we would need to ensure that both the client-side and ledger-side versioning schemes match. If not implemented correctly on the client-side, this could lead to a user inadvertently exercising the wrong choice, which is unacceptable. As discussed before, downside here is that we need to pass a `Claim` object on each exercise, which can be costly; we already do this when originating the instrument though.

The implementation requires that we be able to look up the version # from the `Claim`. For this I've added a new template `AssetDescription.Versions` that has a map `Claim` -> `Version`. In order for this to work, I've had to enable the new LF version to support the new `DA.Map`.

This has the unfortunate side-effect of throwing up a thousand warnings of the old `DA.Next.Map` and `Da.Next.Set`, so I'll need to refactor these if you guys agree with the changes.

This is still WIP; the internal generation of the versions is still not correct. It can only handle simple cases at the moment, so e.g. a bond or european option won't work. The reason for this is that I need a way of generating all the possible combinations of paths the lifecycling can take, deterministically. At the moment I'm using the existing `ContingentClaims.Lifecycle.lifecycle` (previously `settle`) function, which runs the state machine at the given date. This requires me to be able to find all the possible combinations of inputs to this function. So e.g. when I have a `When date == xx/xx/xx` I need to make sure I pass that date in as one of the possible dates. But it's more complicate than this - consider `When date >= xx/xx/xx` or `Cond (observe spotPrice) >= y + x`. In both cases I need to analytically find the fixed point of the function, and then pass that value (two values in the later case - for either combination). So this is not a good approach after all. I'm going to have to write a traversal of the tree from scratch instead - I will focus on this next.

p.s. @GeorgSchneider I considered your suggestion of creating the versions on each lifecycle event triggered by an investor, but I worry this won't work, as the versions will depend on the order the events are triggered in (non-deterministically - consider two investors that make a different choice). So we would need to maintain some sort of a counter state for this to work.


daml/Marketplace/Custody.daml
```diff
@@ -78,7 +79,7 @@ template Service
       nonconsuming RequestLifecycle: ContractId LifecycleRequest
         with
           assetDepositCid: ContractId AssetDeposit
-          postAssetVersion: Int
+          choice: Claims
```
GeorgSchneider, line 82: How do lifecycle event work that don't require an explicit election? Like for a coupon payment, do I have to pass in the "remaining" claims for this, ie. the rhs of the `and` node?

daml.yaml
```diff
@@ -25,3 +25,5 @@ codegen:
   js:
     output-directory: daml.js
     npm-scope: daml.js
+build-options:
+  - --target=1.11
```
brian-weir, line 29: We likely need to discuss the impact of moving to `daml-lf` version 1.11 as it may have impact on the API's calling it (I do agree we should move to it, we just need to ensure we're prepared for it).

Also, does the next version of `daml-lf` help you with your `Claims` work as I believe you had an issue with the size of the contract in the past (?) : 

> Daml-LF 1.12 is now included as an early access feature. Its main improvements are on performance, reducing transaction size through type erasure

daml/Marketplace/Custody.daml
lucianojoublanc-da: Yes, I didn't get around to that yet. I suspect this should be an `Optional`.

daml.yaml
lucianojoublanc-da: This is to support the new `Map`. I explained it in my first comment on the merge page.

daml/Marketplace/Claims.daml
```diff
@@ -21,28 +27,80 @@ assetDescVers: Id
...
+assetDescVers assetId description issuer safekeepingAccountId observers claims = do
+  versionedClaims <- sequence $ ana stepLifecycle (0, fixingDates, claims)
+  assetDescs <- forA versionedClaims \(claims, version) -> create AssetDescription
+                with
+                  assetId = assetId with version
+                  description
+                  issuer
+                  safekeepingAccountId
+                  observers
+                  claims
```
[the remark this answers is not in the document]
lucianojoublanc-da: No - each asset description version still has the entire set of future claims. i.e. the `tail`.

I'm not sure how you could make it a single claim per version. i.e. how would you initialize a `When (at t) x` without an `x`? Something needs to go in the 'hole' to construct the data type.

daml.yaml
brian-weir: Yes, that's not my question - Using the new `Map` has an impact on the calling and receiving APIs and thus we'll need to decide to migrate to it now and update the UI or do it at some later point 

> TextMap itself continues to be supported as there are use-cases where it provides a nicer representation on the API than GenMap. Generic maps are represented as lists of tuples on the APIs. In contrast, TextMap is represented as a JSON object on the JSON API.

lucianojoublanc-da: I do take your point that this may not play nicely e.g. with DamlHub. So we should check that.

lucianojoublanc-da: > Its main improvements are on performance

So I haven't considered this yet. The 1.11 version is for functionality that I need it, not performance.

lucianojoublanc-da: > Yes, that's not my question - Using the new `Map` has an impact on the calling and receiving APIs and thus we'll need to decide to migrate to it now and update the UI or do it at some later point
> 
> > TextMap itself continues to be supported as there are use-cases where it provides a nicer representation on the API than GenMap. Generic maps are represented as lists of tuples on the APIs. In contrast, TextMap is represented as a JSON object on the JSON API.

Hmm - currently **all** `DA.Next.*` are throwing up deprecation warnings, contrary to what the above says. Where did you get that from?

Also, I see what you're saying now, in terms of the JSON map representation. I wasn't aware of that. I don't have an answer at this point.

lucianojoublanc-da: Here's a forum post on this: https://discuss.daml.com/t/clarify-deprecation-of-da-next-in-lf-1-11/2196

I'll now begin migrating `DA.Next`

lucianojoublanc-da: As discussed on slack, we've decided to do this in a separate PR. I will supress the warnings until we do so.

brian-weir (approved):

PR #219, opened 2021-03-02
Matching Engine Bot

brian-weir:


daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -85,7 +88,10 @@ template Service
 
           -- Generate Settlement instructions for allocated bids
           let settleAllocations = (\allocation -> Auction.SettleAllocation with price = finalPrice; issuer = customer; issuerReceivableAccount = receivableAccount; ..) <$> allocatedBids
-          (_, siCids) <- foldlA generateSettlementInstruction (Some depositCid, []) settleAllocations
+          (remainingDepositCidOpt, siCids) <- foldlA generateSettlementInstruction (Some depositCid, []) settleAllocations
+
+          -- Transfer remaining deposit back to the Issuer's trading account
+          whenSome remainingDepositCidOpt \depositCid -> (exerciseByKey @AllocationAccountRule allocationAccount.id Allocation.Withdraw with depositCid = depositCid; transferTo = tradingAccount) $> ()
```
lucianojoublanc-da, line 94: This is the bug I mentioned on slack from ex-bonds wasn't it? :grin:

ui2/package.json
```diff
@@ -18,6 +18,7 @@
     "d3": "^6.5.0",
     "date-fns": "^2.17.0",
     "jsonwebtoken": "^8.5.1",
+    "luxon": "^1.26.0",
```
lucianojoublanc-da, line 21: Think I made same comment last time: do we have some policy for inclusion of libraries? At this rate we're adding a couple every week ... ?

lucianojoublanc-da (approved):
Nice - the matching engine seems pretty concise.

daml/Marketplace/Distribution/Auction/Service.daml
brian-weir: Not sure of the bug in `ex-bonds` (I only checked their dutch auction code) but the purpose of call is if the auction isn't fully filled (ie, there's a deposit existing from the issuer containing units) then return these units to their other normal account from their locked account.

triggers/daml/MatchingEngine.daml
```diff
@@ -0,0 +1,122 @@
...
+import DA.List
+import Daml.Trigger
+
+import qualified Marketplace.Trading.Order as Order
+import qualified Marketplace.Trading.Service as Service
+import qualified Marketplace.Trading.Matching as Matching
+
+import Utils
+
+type CurrentOrderId = Int
```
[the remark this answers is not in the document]
brian-weir: I'd like to take credit for this but it was already their from the previous implementation!

triggers/daml/MatchingEngine.daml
```diff
@@ -0,0 +1,122 @@
...
+  orderCancelRequests <- query @Service.CancelOrderRequest
+  forA_ orderCancelRequests \(cid,_) -> emitExerciseCmd cid Service.AcknowledgeCancel
+                                     >> debug "Acknowledging order cancel"
+
+  -- Check for matches on all 'Order'
+  orders <- query @Order.T
+  forA_ orders (handleOrder party orders)
+
+-- |Check for crossing orders. If found, fill both orders. Matches by price.
+handleOrder : Party -> [ContractPair Order.T] -> ContractPair Order.T -> TriggerA CurrentOrderId ()
```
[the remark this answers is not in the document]
brian-weir: Checking the definition, it looks like its hardcoded to return `()`

> -- | This is the type of your trigger. `s` is the user-defined state type which
-- you can often leave at `()`.
data Trigger s = Trigger
  { initialize : TriggerInitializeA s
  -- ^ Initialize the user-defined state based on the ACS.
  , updateState : Message -> TriggerUpdateA s ()
  -- ^ Update the user-defined state based on a transaction or
  -- completion message.  It can manipulate the state with `get`, `put`,
  -- and `modify`, or query the ACS with `query`.
  **, rule : Party -> TriggerA s ()**
  -- ^ The rule defines the main logic of your trigger. It can send commands
  -- to the ledger using `emitCommands` to change the ACS.
  -- The rule depends on the following arguments:
  --
  --   * The party your trigger is running as.
  --   * The user-defined state.
  --
  -- and can retrieve other data with functions in `TriggerA`:
  --
  --   * The current state of the ACS.
  --   * The current time (UTC in wallclock mode, Unix epoch in static mode)
  --   * The commands in flight.
  , registeredTemplates : RegisteredTemplates
  -- ^ The templates the trigger will receive events for.
  , heartbeat : Optional RelTime
  -- ^ Send a heartbeat message at the given interval.
  }

triggers/daml/MatchingEngine.daml
```diff
@@ -0,0 +1,122 @@
...
+
+      case order.details.orderType of
+        Order.Market -> unless (null oppositelimitOrders') $ fill (orderCid, order) (head oppositelimitOrders')
+        Order.Limit price -> if (not $ null oppositeMarketOrders)
+          then fill (orderCid, order) (head oppositeMarketOrders)
+          else unless (null crossing) $ fill (orderCid, order) (head crossing)
+
+    isOppositeOrder : ContractPair Order.T -> Bool
+    isOppositeOrder (opoCid, opo) = opoCid /= orderCid
+                                 && opo.customer /= order.customer
```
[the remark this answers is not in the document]
brian-weir: This trigger should only be used for testing or demo purposes. The `customer` check here is to stop the same user from matching on both sides of a trade.

ui2/package.json
brian-weir: I've replaced `moment` with `luxon` as `moment` is considered legacy with `luxon` as its replacement. 

As for the policy for inclusion, I haven't seen anything regarding policy but I would expect it should be passing a Blackduck scan to check for vulnerabilities.

PR #222, opened 2021-03-05
Swaption

lucianojoublanc-da:
demo.daml has changed so that we're no longer creating any instruments on startup. Is this intentional?
I've moved the instrument creation into `Demo:demo` form `Demo:issuance`, as the later no longer wrongs (scary script error thrown - I think it's because the `Common` init code was removed).

@georg-da the viz doesn't work because we don't handle the `Give` node. I couldn't figure out how to make it work when I originally implemented the convertible note, it was failing to compile. Maybe you can help me out?


daml/Demo.daml
```diff
@@ -33,6 +33,27 @@ demo = do
 
   issuance providers assets issuer
 
+  let bermudanSwaption floatingRateId fixedRate fixedCcyId principal fixingDates =
```
lucianojoublanc-da, line 36: Try putting this in `issuance`, and then change your `daml.yaml` to start that, and watch the script interpreter suffer :laughing:

daml/Marketplace/Custody.daml
```diff
@@ -311,7 +311,7 @@ template LifecycleRequest
                 (Map.lookup label fixings >>= Map.lookup dt)
               election : Claim Observation Date Id -> Claim Observation Date Id -> Claim Observation Date Id
               election _ _ = deserialize description'.claims -- will throw if the claims don't match
-          payoutCids <- ContingentClaims.settle payout spot election (deserialize description.claims) t
+          payoutCids <- ContingentClaims.lifecycle payout spot election (deserialize description.claims) t
```
lucianojoublanc-da, line 314: Some of my changes in another branch leaked through.

GeorgSchneider (approved):

PR #224, opened 2021-03-09
Remove PublishedBid template and put data on bidder auction

GeorgSchneider:


brian-weir (approved):

daml/Marketplace/Distribution/Bidding/Service.daml
```diff
@@ -56,17 +56,15 @@ template Service
           auctionId : Text
           asset : Asset
           quotedAssetId : Id
+          publishedBidCids : [ContractId Bidding.Bid]
         do
+          publishedBids <- mapA (\bidCid -> do
```
lucianojoublanc-da, line 61: Now that this has changed, it would be cleaner to do:

```haskell
data Bid = Bid with
  public : Public -- I think this overlaps with `Details`
  ... -- all other fields

data Public with
  auctionId, investor, quantity
```

Then this could be written:

```haskell
publishedBids <- forA publishedBidCids \bidCid -> do
   bid <- fetch bidCid
   assert _
   pure bid.public
```
The main point being that now the public part of the bid can be embedded inside the main `Bid`, and we can remove `Details` probably.

PR #227, opened 2021-03-09
Sdk 1.10

alexgraham-da:
Updated to SDK 1.10, added another failsafe step on the `SetupRequired` page, and allowed the exchange to deploy the matching engine trigger during onboarding

### DIFF
--- a/ui/src/components/Investor/OrderForm.tsx
+++ b/ui/src/components/Investor/OrderForm.tsx
@@ -2,6 +2,8 @@ import React, { useState } from 'react'
 import { Button, Form, Header} from 'semantic-ui-react'
 
 import { useParty, useLedger } from '@daml/react'
+import { ContractId } from '@daml/types'
+import { AssetDeposit } from '@daml.js/da-marketplace/lib/DA/Finance/Asset'
 import { Id } from '@daml.js/da-marketplace/lib/DA/Finance/Types/module'
 import { ExchangeParticipant } from '@daml.js/da-marketplace/lib/Marketplace/ExchangeParticipant'
 import { Token } from '@daml.js/da-marketplace/lib/Marketplace/Token'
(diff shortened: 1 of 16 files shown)


ui/src/components/Investor/OrderForm.tsx
```diff
@@ -84,6 +86,7 @@ const OrderForm: React.FC<Props> = ({
     const { ccpMap } = useRegistryLookup();
     const placeOrder = async (kind: OrderKind, deposits: DepositInfo[], amount: string) => {
         const key = wrapDamlTuple([exchange, operator, investor]);
+        const pair = {_1: tokenPair[0], _2: tokenPair[1]};
```
alexgraham-da, line 89: It's unclear to me why `wrapDamlTuple` works with `daml-react` in other places, but not here. I'm guessing it has to do with the fact that it is a tuple of `Id`s, whereas elsewhere it is mostly of strings/parties, but this makes the typechecker happy.

ui/src/components/common/WalletTransaction.tsx
```diff
@@ -126,21 +126,19 @@ const WalletTransaction = (props: {
     )
 
     async function onSubmit() {
+        if (!token || !custodianId) { return; }
         const key = wrapDamlTuple([operator, party]);
-
         setShowSuccessMessage(false)
 
-        let args = {}
```
alexgraham-da, line 133: `daml-react` also got upset when defining `args` up here with no type and using it for two different exercises below.

Sarahb-da: yeah, I have come across this before - it wants all _possible_ values to be the same type 👎  which makes it hard to re-use a variable in this way

Sarahb-da (approved):
looks good!

alexmatson-da (approved):

JudithBLinder (approved):

PR #230, opened 2021-03-15
Dark Mode

brian-weir:


integrationTesting/daml/Exberry.daml
```diff
@@ -51,7 +51,7 @@ submitInstrumentCreation = do
     providerId = "123"
     observers = []
   createListingRequestCid <- alice `submit` do exerciseCmd listingServiceCid L.RequestCreateListing with ..
-  listingCid <- exchange `submit` do exerciseCmd listingServiceCid L.CreateListing with ..
+  -- listingCid <- exchange `submit` do exerciseCmd listingServiceCid L.CreateListing with ..
```
lucianojoublanc-da, line 54: Is this intentional? Don't think it's related to the UI in any way.

lucianojoublanc-da (approved):

integrationTesting/daml/Exberry.daml
brian-weir: It's because I ran an integration test and connected to Exberry Dev and thus wanted to avoid the "auto" creating the listing. Its nothing to do with the UI changes. I left the line in there for testing without explicitly connecting to Exberry on the other side.

PR #231, opened 2021-03-15
Demo fixes

alexgraham-da:
- Add "Clear Market" button for exchange
- Add matching cleared orders to the internal matching engine
- Checkbox for Exchange to deploy matching engine


triggers/daml/MatchingEngine.daml
```diff
@@ -91,3 +140,91 @@ handleOrder party orders (orderCid, order) = do
           timeMatched = show time
 
       return ()
+
+handleClearedOrders : Party -> TriggerA CurrentOrderId ()
```
alexgraham-da, line 144: There is a pretty high amount of copied code here from the `Order` matching (since there's no typeclassing of templates...), I figure it can be cleaned up a bit after the main refactor is merged in.

ui/src/Credentials.ts
```diff
@@ -28,11 +28,11 @@ function isCredentials(credentials: any): credentials is Credentials {
 const CREDENTIALS_STORAGE_KEY = 'credentials';
 
 export function storeCredentials(credentials?: Credentials): void {
-  localStorage.setItem(CREDENTIALS_STORAGE_KEY, JSON.stringify(credentials));
+  sessionStorage.setItem(CREDENTIALS_STORAGE_KEY, JSON.stringify(credentials));
```
alexgraham-da, line 31: The `localStorage` was causing issues demoing where if a tab was left open, it would no longer be refreshable since the credentials were overwritten by the new tab. I think it would make sense to have the DamlHub login use `localStorage`, but for the parties dropdown which is mainly for demoing, use `sessionStorage`.

alexmatson-da: Does `sessionStorage` help here? Data is still stored per origin (domain), so I would expect that having multiple tabs open would still overwrite the selected credentials in storage

alexgraham-da: Yup, it seems to be per individual tab (and I can confirm it does work) -

From https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage:
"Opening multiple tabs/windows with the same URL creates sessionStorage for each tab/window."

alexmatson-da: Nice!

alexmatson-da (approved):

Sarahb-da (approved):

PR #243, opened 2021-03-22
Demo script improvements

GeorgSchneider:
Made some improvements to the demo script:
- Helper functions for several instruments
- Standard setup for instruments with increasing complexity


daml/Marketplace/Issuance/Service.daml
```diff
@@ -69,7 +49,7 @@ template Service
           create ReduceIssuanceRequest with ..
 
     controller provider can
-      nonconsuming Originate: ContractId AssetDescription
+      nonconsuming Originate: (ContractId AssetDescription, AssetDescription)
```
lucianojoublanc-da, line 52: This feels wrong ...

daml/Tests/Utils.daml
```diff
@@ -0,0 +1,66 @@
...
+  in bermudan swap exerciseDates
+
+-- Modeled after https://ledgy.com/blog/examples-of-convertible-notes/
+convertibleNote : Id -> Id -> Decimal -> Decimal -> Date -> Decimal -> Decimal -> C
+convertibleNote underlying ccy principal discount maturity interest cap =
+  let
+    conversionCondition = O.observe underlying.label O.<= O.pure cap
+    conversionPayout = scale (O.pure (principal * (1.0 + interest)) O./ (O.observe underlying.label O.* O.pure (1.0 - discount))) (one underlying)
+    principalPayout = scale (O.pure (principal * (1.0 + interest))) (one ccy)
+  in when (at maturity) $ cond conversionCondition conversionPayout principalPayout
```
lucianojoublanc-da, line 66: As long as we only use these for demos, I don't have an objection (I suspect this is part of your 'building contracts from the ground up' examples). But if these are supposed to be the combinators used for building instruments in production, I think they require some further discussion.

lucianojoublanc-da (approved):

daml/Tests/Utils.daml
GeorgSchneider: Yes, these are purely for demonstration purposes. If we make some of them into combinators and move them into the claims lib, most need generatization.

daml/Marketplace/Issuance/Service.daml
GeorgSchneider: It avoids having to fetch / query after having executed the choice if you need to work with the result, so I think it could be a valid pattern (`fetchByKey` also returns the same). But we can discuss

lucianojoublanc-da: If you're doing this to have the neat list of originations in `Demo.daml`, note that you can achieve something similar by createive use of a list and `<*>`. Something like:

```
[usd, eur] <- sequence $ curry originate <*> [
  ("USD", "United States Dollar", Zero),
  ("EUR", "Euro", Zero)
]
```

lucianojoublanc-da: > It avoids having to fetch / query after having executed the choice if you need to work with the result, so I think it could be a valid pattern (`fetchByKey` also returns the same). But we can discuss

This is what I thought. The downside is that now you can't do the opposite (just the creation), which can be a problem. I'm thinking of a case where you're creating a batch of instruments, say an options chain. This will add the cost of fetching the trees for each option, even if you throw them away afterward. It's especially costly because of the trees, which need to be serialized/deserialized.

PR #245, opened 2021-03-24
Reusable tables & Issuance

alexmatson-da:
This PR starts some work of substituting Material UI components with ones more line with the rest of the styles. I decided to include some simple documentation in this PR description, hopefully you guys will find it helpful when making these substitutions elsewhere! 

### `StripedTable`

Importable from `components/Table/StripedTable`, this is (one of) the component that we used in `ui` on master for displaying tables.  We should be able to convert all the existing Material UI tables to StripedTables pretty easily. The table headers and rows are passed in as props. 

See `Instruments.tsx` in origination (in this PR) for an example. 

### `Tile`

Tile can be used for blocks of content on a page. Pretty self-explanatory, it takes in children for content but also has a `header` prop for a title. 

### `FormErrorHandled`

This is a component that we wrapped around the Semantic UI `Form` in the old codebase, which has built-in reusable functionality for dealing with any errors thrown inside the `onSubmit` callback, and displays an error toast message on screen. Also automatically handles `loading` states if the `onSubmit` callback is async. 

In the children, use the typical [Semantic UI form components](https://react.semantic-ui.com/collections/form/). 

An example of usage in this PR can be found in `pages/issuance/New.tsx`. 

# Comparison shots!

Before
Screen Shot 2021-03-24 at 3 46 34 PM

After
Screen Shot 2021-03-24 at 3 41 52 PM

### DIFF
--- a/ui2/src/pages/origination/Instruments.tsx
+++ b/ui2/src/pages/origination/Instruments.tsx
@@ -1,71 +1,49 @@
 import React from "react";
 import { withRouter, RouteComponentProps } from "react-router-dom";
-import { Table, TableBody, TableCell, TableRow, TableHead, Button, Grid, Paper, Typography } from "@material-ui/core";
 import { IconButton } from "@material-ui/core";
 import { KeyboardArrowRight } from "@material-ui/icons";
 import { useStreamQueries } from "@daml/react";
-import useStyles from "../styles";
 import { getName } from "../../config";
 import { AssetDescription } from "@daml.js/da-marketplace/lib/Marketplace/Issuance/AssetDescription";
+import StripedTable from "../../components/Table/StripedTable";
+import { Button } from "semantic-ui-react";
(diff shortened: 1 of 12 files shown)


brian-weir (approved):

brian-weir:
Thanks @alexmatson-da for the PR and the nice explanation. Much appreciated!!

I hope you don't mind but I merged the PR to sales-eng so I can already start migrating to these components this morning.

ui2/src/pages/origination/Instruments.tsx
```diff
@@ -1,72 +1,50 @@
...
-                    </TableRow>
-                  ))}
-                </TableBody>
-              </Table>
-            </Paper>
-          </Grid>
-        </Grid>
-      </Grid>
-    </>
+    <div className='instruments'>
```
brian-weir, line 17: Where is this class defined ?

PR #248, opened 2021-03-29
Setup widget

Sarahb-da:


alexmatson-da (approved):

ui/src/components/QuickSetup.tsx
```diff
@@ -0,0 +1,598 @@
...
+interface IPartyLoginData extends PartyDetails {
+    role: MarketRole
+    name: string
+    location: string
+    title?: string
+    issuerId?: string
+    deployMatchingEngine?: boolean
+}
+
+const MARKETROLES = ["CustodianRole", "IssuerRole", "ExchangeRole", "InvestorRole", "BrokerRole"]
```
alexgraham-da, line 50: I think the `MarketRole` enum values would work here directly as well (they type just fine to `string`).

Sarahb-da: oh thanks!

alexgraham-da (approved):
The setup widget is missing the CCP role currently - not sure if by design. It's a bit of a special snowflake like Exchange and Issuer since it needs a dropdown of custodians, so we can always add it later if you need to get this out of the door.

Otherwise this looks good. Very nifty!

PR #253, opened 2021-03-31
Add registry table

Sarahb-da:


ui/src/components/QuickSetup.tsx
```diff
@@ -203,6 +227,95 @@ const QuickSetup = (props: { onLogin: (credentials?: Credentials) => void }) =>
...
+}
+
+const RegistryTableRow = (props: { index: number; party: PartyDetails; roles: string[] }) => {
+    const { index, party, roles } = props
+    const oddRow = index % 2 === 0 ? "odd-row" : undefined
+    const partyName = <p className='bold'>{party.partyName}</p>
+
+    if (roles.length === 0) {
+        return (
+            <Table.Row className={oddRow}>
```
alexmatson-da, line 301: Looking just at the code, I think there may be a chance that this ends up with `<tr class='undefined'>` in the final page HTML if oddRow is undefined.

It'd be safer to have the ternary on 296 say `"odd-row" : ""` instead, in the case of an even row

alexmatson-da (approved):
cool!

alexgraham-da (approved):
looks good!

PR #259, opened 2021-04-07
Add landing page and network request modals

alexmatson-da:


ui2/src/pages/landing/Landing.tsx
```diff
@@ -0,0 +1,191 @@
...
+import { Request as TradingRequest } from "@daml.js/da-marketplace/lib/Marketplace/Trading/Service/module";
+import { VerifiedIdentity } from '@daml.js/da-marketplace/lib/Marketplace/Regulator/Model';
+import { Template } from '@daml/types';
+import { AssetSettlementRule } from '@daml.js/da-marketplace/lib/DA/Finance/Asset/Settlement';
+import { Account } from '@daml.js/da-marketplace/lib/DA/Finance/Types';
+
+
+function hashUserName(name: string): number {
+  // Hash a user name to map to values in the range [1, 4], to determine profile pic color
+  return (name.split('').reduce((sum, char) => sum + char.charCodeAt(0), 0) % 4) + 1;
```
Sarahb-da, line 27: nice

ui2/src/icons/icons.tsx
```diff
@@ -17,6 +17,18 @@ export const ArrowRightIcon = () => (
     </svg>
 )
 
+export const CogIcon = () => (
+    <svg width="19" height="20" viewBox="0 0 19 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Sarahb-da, line 21: just for the sake of following the pattern here - `className="icon cog-icon"` and you probably only need the viewbox

ui2/src/pages/page/TopMenu.tsx
```diff
@@ -30,7 +30,7 @@ const TopMenu: React.FC<Props> = ({ title, buttons, activeMenuTitle, buttonDivid
         <div className='top-section'>
             <Menu className='top-menu'>
                 <Menu.Menu position='left'>
-                    <Menu.Item disabled={!activeMenuTitle} onClick={history.goBack}>
+                    <Menu.Item as={!activeMenuTitle ? 'div' : undefined} disabled={!activeMenuTitle} onClick={history.goBack}>
```
Sarahb-da, line 33: does this work correctly? if activeMenuTitle is false, display the Item but disable it - otherwise display nothing? 

I could be misinterpreting what `as` is doing here

Sarahb-da (approved):
one comment but overall looks great

ui2/src/context/ServicesContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+    throw new Error("useProviderServices  must be used within a ServicesProvider");
+  }
+
+  return context.services
+    .filter(s => s.contract.payload.customer === party)
+    .reduce((acc, service) => {
+      const providerDetails = acc.find(i => i.provider === service.contract.payload.provider);
+
+      const provider = providerDetails?.provider || service.contract.payload.provider;
+      const services = [ ...(providerDetails?.services || []), service.service];
```
brian-weir, line 111: You could use `_.compact(providerDetails.services)` from lodash which matches the pattern above where you return an empty list if an object is undefined, null, etc.

See https://lodash.com/docs/4.17.15#compact.

brian-weir (approved):

ui2/src/pages/page/TopMenu.tsx
alexmatson-da: By default, `Menu.Item` renders as a NavLink (eventually, an `<a>` tag) which gets disabled when `!activeMenuTitle`, but is still an `<a>`. 

The issue is that the Welcome Header has `<a>` tags as well, that link to Daml and Daml Hub. So the browser console warns that there are nested `<a>` tags in the DOM, which is invalid HTML

So with `as={!activeMenuTitle ? 'div' : undefined}`, if activeMenuTitle is false then display this menu item as a `<div>` (which in `ui2`, activeMenuTitle is actually always false. I haven't used that prop yet). 

If activeMenuTitle is enabled, the undefined will make it a NavLink by default, which makes sense in this case because it's a link to goBack (we don't want to break the accessibility/browser behavior that comes along with _real_ links vs interactive divs)

ui2/src/context/ServicesContext.tsx
alexmatson-da: This is awesome, thanks!

PR #264, opened 2021-04-08
fix binary option test

richardkapolnai-da:


daml/Tests/BinaryOption.daml
```diff
@@ -25,13 +28,14 @@ endToEnd = script do
...
-      underlyingAsset = Asset with id = underlyingId, quantity = 200.0
+      strike = 910.0: Decimal
+      expiration = date 2020 Mar 18
 
   alice <- onboardCustomer providers "Alice"
   bob <- onboardCustomer providers "Bob"
 
   -- Alice originates the option
-  let boClaims = binaryCallOption underlyingId usdId 700.0 (date 2022 Oct 22)
+  let boClaims = binaryCallOption underlyingId usdId strike expiration
```
richardkapolnai-da, line 38: I believe this was unintentional, hopefully fixed now.

daml/Tests/BinaryOption.daml
```diff
@@ -46,7 +50,7 @@ endToEnd = script do
       issuanceId = "Myissuance"
       accountId = Id (Set.fromList [alice.customer, providers.bank]) alice.mainAccount.id.label 0
       assetId = optionDesc.assetId
-      quantity = 1000.0
+      quantity = 1000.0 -- TODO is this unused ???
```
richardkapolnai-da, line 53: Apparently this number does not influence anything, is this a bug?

daml/Tests/BinaryOption.daml
lucianojoublanc-da: In both cases this was intentional.

The price is different because the payoff/cashflow of this instrument during settlement is the `strike - price`, so if they were the same you would settle 0$.

The date (not just the month) was different in order to avoid triggering the lifecycle. This was because there was a bug in the lifecycle process at the time, so I wanted to avoid it triggering. I guess I just forgot to put it back to the expiry date. It should be safe to do now.

daml/Tests/BinaryOption.daml
lucianojoublanc-da: This is used to create the `Issuance` contract. That is, how many options are issued and created with the `AssetDeposit`. It is different from the payoff (which is 1$).

daml/Tests/BinaryOption.daml
```diff
@@ -72,4 +76,11 @@ endToEnd = script do
   submit providers.bank $ exerciseCmd alice.custodyServiceCid $ Lifecycle with
     safekeepingDepositCid = corpActDepositCid -- TODO: how do we get this in practice?
     lifecycleRequestCid = lifecycleReqCid
-    fixings = Map.fromList [ ("TSLA", Map.insert (date 2020 Mar 18) 920.0 Map.empty) ]
+    fixings = Map.fromList [ ("TSLA", Map.insert expiration 920.0 Map.empty) ]
+
+  [(_, bobsOption)] <- queryFilter @AssetDeposit bob.customer (\c -> c.asset.id.label == "TSLA Mar21 910.0")
+  bobsOption.asset.id.version === 1
+  [(_, bobsMoney)] <- queryFilter @AssetDeposit bob.customer (\c -> c.asset.id.label == "USD")
+  bobsMoney.asset.quantity === 1.0 -- TODO why not 1000.0, as Bob has 1000 options (quantity = 1000.0)
```
lucianojoublanc-da, line 84: As per your comment, I would expect there to be an asset deposit for 1000$. Are these tests passing?

daml/Tests/BinaryOption.daml
richardkapolnai-da: Isn't the price 920 at expiry (see line 75)?
The assetLabel (line 40) has the strike, doesn't it?

daml/Tests/BinaryOption.daml
richardkapolnai-da: It still seems surprising that bob had 1000 binary options, they are _all_ exercised, and he gets only $1.
Then they shouldn't be all exercised perhaps?

lucianojoublanc-da: Yes, I see what you mean now. I made a minor change on this branch to fix it. I hope you don't mind.

daml/Tests/BinaryOption.daml
lucianojoublanc-da: > Was this unintentionally different (700 vs 910, Oct vs Mar)?
Yes, I see what you mean now. I thought you were comparing the prices and expiry to that of the strike/ledger date. I didn't realise you meant the `assetLabel`. So rather yes, the `assetLabel` is incorrect.

Ideally we should just generate the `assetLabel` programmatically below to ensure it matches the option parameters.

daml/Tests/BinaryOption.daml
```diff
@@ -20,18 +23,19 @@ endToEnd = script do
...
+      expiration = date 2020 Mar 18
 
   alice <- onboardCustomer providers "Alice"
   bob <- onboardCustomer providers "Bob"
 
   -- Alice originates the option
-  let boClaims = binaryCallOption underlyingId usdId 700.0 (date 2022 Oct 22)
+  let boClaims = binaryCallOption underlyingId usdId strike expiration
   origReqCid <- submit alice.customer $ exerciseCmd alice.issuanceServiceCid RequestOrigination with
       assetLabel = "TSLA Mar21 910.0"
```
lucianojoublanc-da, line 40: Something like (untested):

```suggestion
      assetLabel = words [underlyingId.label, show expiration.month, expiration.day, strike]
```

richardkapolnai-da: I think the handcrafted string is much easier to read, but we can do some wizardry if you prefer

daml/Tests/BinaryOption.daml
richardkapolnai-da: (fixed)

PR #266, opened 2021-04-13
New auto approval trigger

brian-weir:


triggers/daml/AutoApproval.daml
```diff
@@ -0,0 +1,132 @@
...
+    matchingserviceOffers ++
+    matchingserviceRequests
+
+-- | Facilitates querying a template by a specific field which matches a provided filter predicate
+queryByField : forall t f a m. (Template t, HasField f t a, ActionTriggerAny m, Functor m) => (t -> a) -> (a -> Bool) -> m [(ContractId t)]
+queryByField getField pred = map fst . filter (pred . getField . snd) <$> query
+
+-- | Generates commands in the expected format expected by the trigger for a contract / choice combination
+generateCommands : forall t c r. (Template t, Choice t c r) => c -> [(ContractId t)] -> [(AnyContractId, Command)]
+generateCommands choiceArg queryResults = foldl (\acc cid -> (toAnyContractId cid, exerciseCmd cid choiceArg) :: acc) [] queryResults
```
alexgraham-da, line 117: I'm curious, why use `foldl` here instead of `map`, unless you prefer the resulting list to be in reverse?

brian-weir: `foldl` makes the code look more complex and thus it gives the general impression that I know what I'm actually doing 😄 

Joking aside, the reason is that initially this function was much more complicated that warranted a `foldl` but since I've removed that logic, I could have also replaced the `foldl` with a `map` to make the entire function simpler. I'll update it in the next PR.

alexgraham-da: Haha fair enough! Related, I can't tell you how many times I've written what I thought was a beautiful and elegant recursive function that I realize later could have just been a simple fold...

daml/Marketplace/Custody/Role.daml
```diff
@@ -39,6 +39,12 @@ template Role
         do
           create Issuance.Offer with ..
 
+      nonconsuming AcceptIssuanceRequest : ContractId Issuance.Service
```
lucianojoublanc-da, line 42: I'm super confused - how could we have a `OfferIssuanceService` previously, but not an `Accept*`?

daml/Marketplace/Distribution/Role.daml
```diff
@@ -20,6 +20,12 @@ template Role
         do
           create Auction.Offer with ..
 
+      nonconsuming ApproveAuctionServiceRequest : ContractId Auction.Service
```
lucianojoublanc-da, line 23: Same Q as in my previous comment. Didn't we have choices to approve auction requests previuosly? If not, how was it done? Or have you removed some other choice? I can't see one though.

triggers/daml/AutoApproval.daml
```diff
@@ -0,0 +1,132 @@
...
+  , heartbeat = None
+  }
+
+-- | Approval all requests
+handleApprovalRule : Party -> TriggerA () ()
+handleApprovalRule party = do
+  debug "Running Auto Approval rule..."
+
+  -- Exercise Accept/Approve choice on the triggered contract
+  custodyRoleOffers         <- generateCommands CustodyRole.Accept        <$> queryByField @CustodyRole.Offer         @"provider" (.provider) (== party)
```
lucianojoublanc-da, line 57: Are you not able to omit the type annotation `@provider` here? why isn't it inferred from `(.provider)`?

lucianojoublanc-da: Also, as you're using `(== party)` in every call, I would consider removing that argument, rename `queryFilterParty` or something like that.

triggers/daml/AutoApproval.daml
```diff
@@ -0,0 +1,132 @@
...
+  custodyServicesRequests   <- queryByField @CustodyService.Request  @"provider" (.provider) (== party)
+  issuanceServiceRequests   <- queryByField @IssuanceService.Request @"provider" (.provider) (== party)
+  custodyRoles              <- queryByField @CustodyRole.Role        @"provider" (.provider) (== party)
+  let custodyRequests  = generateRoleCommands (\cid -> CustodyRole.AcceptCustodyRequest with custodyRequestCid = cid) custodyServicesRequests custodyRoles
+      issuanceRequests = generateRoleCommands (\cid -> CustodyRole.AcceptIssuanceRequest with issuanceServiceRequestCid = cid) issuanceServiceRequests custodyRoles
+
+  auctionServiceRequests    <- queryByField @AuctionService.Request @"provider" (.provider) (== party)
+  biddingServiceRequests    <- queryByField @BiddingService.Request @"provider" (.provider) (== party)
+  distributionRole          <- queryByField @DistributionRole.Role  @"provider" (.provider) (== party)
+  let auctionRequests = generateRoleCommands (\cid -> DistributionRole.ApproveAuctionServiceRequest with auctionServiceRequestCid = cid) auctionServiceRequests distributionRole
```
lucianojoublanc-da, line 82:
```suggestion
  let auctionRequests = generateRoleCommands DistributionRole.ApproveAuctionServiceRequest auctionServiceRequests distributionRole
```

lucianojoublanc-da (commented):
Looks good; I'd consider refactoring `queryByField` to be more specific; it will cut down code and improve legibility. I can see why you did this, i.e. you didn't want to repeat yourself, but I think you can cut down the code considerably by doing away with some abstractions, see my comments.

daml/Marketplace/Custody/Role.daml
brian-weir: The access pattern we use is the following :

1. A **Provider** can `Offer` a service where the responsibility lies with the **Customer** to either `Accept` or `Reject` said `Offer`. In most cases, this requires the signature of both the **Operator** and the **Provider** and thusly is why a `Role` contract with both signatures is required to make such an `Offer` ;
2. A **Customer** can make a `Request` for a given service where the responsibility lies with the **Provider** (and sometimes the **Operator**) to `Approve` the said `Request` 

For example, you can see this pattern by checking the required signatories and parties on controllers on the `Offer` and `Request` contracts for the `Custody` service  :

>     template Offer
>       with
>         operator : Party
>         provider : Party
>         customer : Party
>       where
>         signatory operator, provider
> 
>     controller customer can
>       Accept : ContractId Service
>         do
>           create Service with ..
> 
>       Decline : ()
>         do pure ()
> 
>     controller provider can
>       Withdraw : ()
>         do pure ()
> 
>     template Request
>       with
>         customer : Party
>         provider : Party
>       where
>         signatory customer
>         observer provider
> 
>     controller customer can
>       Cancel : ()
>         do pure ()
> 
>     controller provider can
>       Reject : ()
>         do pure ()
> 
>     choice Approve : ContractId Service
>       with
>         operator : Party
>       controller operator, provider
>       do
>         create Service with ..

This pattern is also used for role contracts where the **Operator** `Offer`s roles and correspondingly the **Provider** `Request`s roles where the responsibility lies with the relating party to `Accept` an `Offer` or `Approve` a `Request`.

lucianojoublanc-da: I understand the pattern, but I don't understand why you've had to add the `Accept` choice _now_, in this PR. Did not exist in `sales-eng` already? Otherwise, how could the existing `Request` be completed (line 40)?

triggers/daml/AutoApproval.daml
brian-weir: To your second point, you're correct - I had also through of refactoring this but left it since this trigger is a `catch all` for accept/approval of any contract which we may require other filtering other than `Party` in the future. 

I'll replace this with your suggestion since we currently don't know what these future requirements are and currently we only need to filter on `Party`

triggers/daml/AutoApproval.daml
brian-weir: Good catch - I'll apply this to the other relating variables too!

brian-weir (commented):
> Looks good; I'd consider refactoring `queryByField` to be more specific; it will cut down code and improve legibility. I can see why you did this, i.e. you didn't want to repeat yourself, but I think you can cut down the code considerably by doing away with some abstractions, see my comments.

Thanks for your feedback! Much appreciated. I'll refactor the code and also take in @alexgraham-da's feedback in this PR.

daml/Marketplace/Custody/Role.daml
brian-weir: I added it now as I only noticed now that it was missing - the demo setup is using the `Offer` contract and thus I didn't realise it was missing.

triggers/daml/AutoApproval.daml
brian-weir: Argh, I could have levelled up in complexity by writing my own custom recursive function. I must remember that for next time...

In the meantime, I've refactored out `foldl` with `map` on the last commit 👍

lucianojoublanc-da (approved):

PR #271, opened 2021-04-15
Quick setup services

Sarahb-da:


ui2/src/pages/login/QuickSetup.tsx
```diff
@@ -0,0 +1,560 @@
...
+    if (!operatorService.loading && operatorService.contracts.length === 0) {
+      createOperatorService();
+    }
+  }, [operatorService.loading, operatorService]);
+
+  const partyOptions = parties.map(party => {
+    return { text: party.partyName, value: party.party };
+  });
+
+  let serviceOptions = services.map(service => {
```
alexmatson-da, line 160: you can map over the enum directly, and avoid having to maintain two different vars with service lists in them.

`Object.values(ServiceKind).map(...)` should work, (after importing ServiceKind -- see above), and now we don't need the `services` array variable

ui2/src/pages/login/QuickSetup.tsx
```diff
@@ -0,0 +1,560 @@
...
+      return;
+    }
+
+    let newMarketData: Map<string, string[]> = new Map();
+
+    custodianRoles.contracts.forEach(c => addMarketData(c.payload.provider, 'Custody'));
+    distributorRoles.contracts.forEach(c => addMarketData(c.payload.provider, 'Listing'));
+    settlementServices.contracts.forEach(c => addMarketData(c.payload.provider, 'Settlement'));
+    exchangeRoles.contracts.forEach(c => addMarketData(c.payload.provider, 'Trading'));
+    matchingServices.contracts.forEach(c => addMarketData(c.payload.provider, 'Matching'));
```
alexmatson-da, line 130: we should use the enum values instead of string literals, to keep things in one spot

ui2/src/pages/login/QuickSetup.tsx
```diff
@@ -0,0 +1,560 @@
...
+
+type Offer = CustodianOffer | DistributorOffer | SettlementOffer | ExchangeOffer | MatchingOffer;
+
+enum serviceOptionsEnum {
+  CUSTODY = 'Custody',
+  TRADING = 'Trading',
+  MATCHING = 'Matching',
+  SETTLEMENT = 'Settlement',
+  LISTING = 'Listing',
+}
```
alexmatson-da, line 58: I defined these data types in my recent PR, would be good to switch over! 

`ServiceOffer` imported from `context/ServicesContext.tsx` and 

`ServiceKind` imported from the same file. It is also a service string enum 

You'll have to add some additional services (e.g., Matching) in those though

ui2/src/pages/login/Login.tsx
```diff
@@ -106,6 +112,26 @@ const LoginScreen: React.FC<Props> = ({ onLogin }) => {
...
+      setParties(parties);
+    }
+  }, []);
+
+  return (
+    <>
+      <Button className="ghost dark" onClick={() => history.push('/quick-setup')}>
+        Quick Setup
+      </Button>
+    </>
```
alexmatson-da, line 131: redundant outer fragment `<> ... </>` can be removed

alexmatson-da: also should probably be a `Link` or `NavLink` instead :D 

if you add the `'button ghost dark'` className, I _think_ it will be styled to look like a button...

ui2/src/pages/login/QuickSetup.tsx
alexgraham-da: I was going to use this trick for `MarketRole` (the daml sum type enum) for something but the Daml codegen adds a few extra entries related to the deserialization so it wouldn't have worked 😞

alexgraham-da (approved):
looks good to me! now that we have the autoaccept trigger (and I'm guessing there will be a few more added), it would be nice for this page to also allow the user to specify which (if any) triggers to start up for the parties they're managing. I just added the clearing service and trigger for that, so I could try adding that in and also add clearing to this in a separate PR

ui2/src/pages/login/Login.tsx
Sarahb-da: unfortunately that does not style - i think it could if we added some more foundational button styling but will save for different PR

ui2/src/pages/login/QuickSetup.tsx
Sarahb-da: ah yes thanks

ui2/src/pages/login/Login.tsx
alexmatson-da: it is unfortunate due to the loss of the keyboard accessibility, but i do not think that it is urgent enough to block the rest of this PR. approved :D

alexmatson-da (approved):

PR #278, opened 2021-04-20
Sales eng clearing setup

alexgraham-da:
add clearing setup

### DIFF
--- a/daml/Marketplace/Clearing/Service.daml
+++ b/daml/Marketplace/Clearing/Service.daml
@@ -170,7 +170,10 @@ template Service
         ctrl : Party
       controller ctrl
       do
-        void $ archive . fst <$> fetchByKey @MemberStanding (provider,customer)
+        (standingCid,_) <- fetchByKey @MemberStanding (provider,customer)
+        debug standingCid
+        archive standingCid
+
 
 template Offer
   with
--- a/daml/Marketplace/Clearing/Service.daml
+++ b/daml/Marketplace/Clearing/Service.daml
@@ -171,7 +171,6 @@ template Service
       controller ctrl
       do
         (standingCid,_) <- fetchByKey @MemberStanding (provider,customer)
-        debug standingCid
         archive standingCid
 
 
--- a/ui2/src/pages/common.ts
+++ b/ui2/src/pages/common.ts
@@ -1,5 +1,30 @@
 import { CreateEvent } from '@daml/ledger';
+import {DropdownProps, DropdownItemProps} from 'semantic-ui-react';
 
 export type ServicePageProps<T extends object> = {
   services: Readonly<CreateEvent<T, any, any>[]>;
 };
+
+export function isStringArray(strArr: any): strArr is string[] {
+    if (Array.isArray(strArr)) {
+        return strArr.reduce((acc, elem) => {
+            return acc && typeof elem === 'string'
+        }, true);
+    } else {
+        return false
+    }
+}
+
+export const handleSelectMultiple = (result: DropdownProps, current: string[], setter: React.Dispatch<React.SetStateAction<string[]>>) => {
+    if (typeof result.value === 'string') {
+        setter([...current, result.value])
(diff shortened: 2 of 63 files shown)


ui2/src/pages/network/Clearing.tsx
```diff
@@ -0,0 +1,218 @@
...
+  const rejectOffer = async (c: CreateEvent<Offer>) => {
+    await ledger.exercise(Offer.Decline, c.contractId, {});
+  };
+
+  return (
+    <div className="assets">
+      <Header as="h3">Current Services</Header>
+      <StripedTable
+        headings={['Service', 'Operator', 'Provider', 'Consumer', 'Role', 'Action' /* 'Details' */]}
+        rows={services.map((c, i) => {
```
Sarahb-da, line 93: there is a loading prop for striped table now, if interested

ui2/src/pages/network/Clearing.tsx
```diff
@@ -0,0 +1,218 @@
...
+              >
+                Terminate
+              </Button>,
+            ],
+          };
+        })}
+      />
+      <Header as="h3">Requests</Header>
+      <StripedTable
+        headings={['Type', 'Consumer', 'Actions' /* 'Details' */]}
```
Sarahb-da, line 115: can use the loading prop here too

ui2/src/pages/landing/Landing.tsx
```diff
@@ -61,6 +63,9 @@ interface RequestInterface {
   provider: string;
   tradingAccount?: Account;
   allocationAccount?: Account;
+
```
Sarahb-da, line 66: extra line

Sarahb-da (approved):
👍

ui2/src/pages/network/Clearing.tsx
alexgraham-da: nice 👍

alexmatson-da (approved):

daml/Marketplace/Clearing/Service.daml
```diff
@@ -170,7 +170,9 @@ template Service
         ctrl : Party
       controller ctrl
       do
-        void $ archive . fst <$> fetchByKey @MemberStanding (provider,customer)
+        (standingCid,_) <- fetchByKey @MemberStanding (provider,customer)
```
lucianojoublanc-da, line 173: This is ... much easier to understand! :smile:

ui2/src/pages/common.ts
```diff
@@ -1,5 +1,38 @@
 import { CreateEvent } from '@daml/ledger';
+import { DropdownProps, DropdownItemProps } from 'semantic-ui-react';
 
 export type ServicePageProps<T extends object> = {
   services: Readonly<CreateEvent<T, any, any>[]>;
 };
+
+export function isStringArray(strArr: any): strArr is string[] {
```
lucianojoublanc-da, line 8: My Daml mind rebels against this. Is there a way to refactor the code so you're not passing `any` around, to avoid using this function?

daml/Marketplace/Clearing/Service.daml
alexgraham-da: Weirdly enough, doing it the original way didn’t actually end up archiving the contract...which makes zero sense to me. So it’s possible I had changed something else between it working and it not working.

ui2/src/pages/common.ts
alexgraham-da: This was copied over in `ui`, to have a select dropdown that can also function as a regular text input...not sure it's actually necessary with the type of dropdowns I'm using it for, so I can take a look at getting rid of it.

I also don't enjoy seeing `any`, but using it for a function that checks what a type is (such as `Array.isArray`) does make sense to me - especially because it's use here is to make dealing with `DropdownProps` more "type safe", since `DropdownProps` is `[key: string]: any` which there's not much you can do about.

PR #284, opened 2021-04-21
Regulator improvements

alexmatson-da:
Some improvements to the regulator / identity stuff throughout the UI

### DIFF
--- a/triggers/daml/AutoApproval.daml
+++ b/triggers/daml/AutoApproval.daml
@@ -34,6 +34,7 @@ autoApprovalTrigger = Trigger
     , registeredTemplate @ListingService.Offer
     , registeredTemplate @RegulatorService.Offer
     , registeredTemplate @RegulatorService.Request -- TO-DO: Auto-approve requests
+    , registeredTemplate @RegulatorService.Service
     , registeredTemplate @RegulatorService.IdentityVerificationRequest
     , registeredTemplate @SettlementService.Offer
     , registeredTemplate @SettlementService.Request
--- a/triggers/daml/AutoApproval.daml
+++ b/triggers/daml/AutoApproval.daml
@@ -14,6 +14,7 @@ import Marketplace.Distribution.Bidding.Service qualified as BiddingService
 import Marketplace.Issuance.Service qualified as IssuanceService
 import Marketplace.Listing.Service qualified as ListingService
 import Marketplace.Regulator.Service qualified as RegulatorService
+import Marketplace.Regulator.Model qualified as Regulator
 import Marketplace.Settlement.Service qualified as SettlementService
 import Marketplace.Trading.Role qualified as TradingRole
 import Marketplace.Trading.Service qualified as TradingService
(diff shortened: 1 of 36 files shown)


triggers/daml/AutoApproval.daml
```diff
@@ -134,6 +137,11 @@ handleApprovalRule party = do
     generateServiceCommands IssuanceService.Originate      originationRequests issuanceServices ++
     generateServiceCommands IssuanceService.CreateIssuance issuanceRequests issuanceServices
 
+  idVerificationRequests  <- queryByParty @RegulatorService.IdentityVerificationRequest (.provider)
+  regulatorServices       <- queryByParty @RegulatorService.Service              (.provider)
+  submitCommands $
+    generateServiceCommands RegulatorService.VerifyIdentity idVerificationRequests regulatorServices
+
   creditRequests          <- queryByParty @Custody.CreditAccountRequest (.provider)
```
alexmatson-da, line 145: This builds, but it doesn't do anything when an `IdentityVerificationRequest` comes in. @brian-weir is it obvious why that might be from this code?

ui2/src/pages/landing/Landing.tsx
```diff
@@ -90,21 +112,28 @@ const ProfileSection: React.FC<{ action: () => void }> = ({ action }) => {
   const { contracts: identities, loading: identitiesLoading } = useStreamQueries(VerifiedIdentity);
   const partyIdentity = identities.find(id => id.payload.customer === customer);
 
+  const damlHubParty =
+    isDamlHubParty(customer) && name !== customer ? (
+      <p>
```
Sarahb-da, line 117: would it also make sense to put the `<p>` on line 46 rather than potentially wrapping an `input` in a `<p>` (if its a daml hub party)

Sarahb-da (approved):
👍

triggers/daml/AutoApproval.daml
brian-weir: It looks like you're missing `registeredTemplate @RegulatorService.Service` as part of the `registeredTemplates` in the trigger definition - meaning we won't receive events for this template and thus we can't query on it. 

Therefore, `regulatorServices` will always be an empty list and when you call `generateServiceCommands` with this empty list, this in turn calls `generateCommandsWithCid` which matches on `generateCommandsWithCid _ _ _ [] = []`. Then when `submitCommands' is called with an empty list it will match on `submitCommands [] = pure ()` and thus no side effects are executed on the ledger.

I understand that's a lot to take in, we can discuss further if you wish.

brian-weir (approved):

alexgraham-da (approved):

triggers/daml/AutoApproval.daml
alexgraham-da: I've been caught by this as well, especially when first starting writing triggers coming from DAZL bots - you need to add a `registeredTemplate` for both contracts you want to cause the trigger to run, _and_ any contract you need to query otherwise.

alexmatson-da: Thanks, that was very helpful to understanding my problem and also how the trigger works in general! After registering the Service template, the Operator now auto-approves verified identity requests successfully. 

I'd still like to find a way to auto-approve the Regulator service requests themselves. Will you available after today's catch up to go through it with me?

ui2/src/pages/landing/Landing.tsx
alexmatson-da: Thanks, yeah that makes more sense. I also realized that since I'm checking `isDamlHubParty` out here (line 115), I don't need to do the same thing inside the component itself

PR #286, opened 2021-04-23
Minor fixes

brian-weir:


triggers/daml/AutoApproval.daml
```diff
@@ -167,7 +167,11 @@ generateCommands choiceArg queryResults = map (\(cid, _) -> (toAnyContractId cid
 generateCommandsWithCid : forall t c r t2. (Template t, Choice t c r, Template t2) => (t2 -> t -> Bool) -> (ContractId t2 -> c) -> [(ContractId t2, t2)] -> [(ContractId t, t)] -> [(AnyContractId, Command)]
 generateCommandsWithCid _ _ [] _ = []
 generateCommandsWithCid _ _ _ [] = []
-generateCommandsWithCid pred toChoice requests targets = mapOptional (\(reqCid, req) -> (\(cid, _) -> (toAnyContractId cid, exerciseCmd cid $ toChoice reqCid)) <$> findFirstMatch req targets pred) requests
+generateCommandsWithCid pred toChoice requests targets = mapOptional (\(reqCid, req) -> getCommandWithCid toChoice reqCid <$> findFirstMatch req targets pred) requests
+
+-- | Takes a choice and a contract id which is a parameter of that choice, the contract which exposes this choice and builds up a command
+getCommandWithCid : forall t c r t2. (Template t, Choice t c r, Template t2) => (ContractId t2 -> c) -> ContractId t2 -> (ContractId t, t) -> (AnyContractId, Command)
+getCommandWithCid toChoice requestCid (cid, _) = (toAnyContractId cid, exerciseCmd cid $ toChoice requestCid)
```
brian-weir, line 174: The goal of doing this is to make it easier to comprehend what this function is doing, by removing one of the lambda's out into its own function.

triggers/daml/AutoApproval.daml
```diff
@@ -167,7 +167,11 @@ generateCommands choiceArg queryResults = map (\(cid, _) -> (toAnyContractId cid
 generateCommandsWithCid : forall t c r t2. (Template t, Choice t c r, Template t2) => (t2 -> t -> Bool) -> (ContractId t2 -> c) -> [(ContractId t2, t2)] -> [(ContractId t, t)] -> [(AnyContractId, Command)]
 generateCommandsWithCid _ _ [] _ = []
 generateCommandsWithCid _ _ _ [] = []
-generateCommandsWithCid pred toChoice requests targets = mapOptional (\(reqCid, req) -> (\(cid, _) -> (toAnyContractId cid, exerciseCmd cid $ toChoice reqCid)) <$> findFirstMatch req targets pred) requests
+generateCommandsWithCid pred toChoice requests targets = mapOptional (\(reqCid, req) -> getCommandWithCid toChoice reqCid <$> findFirstMatch req targets pred) requests
+
+-- | Takes a choice and a contract id which is a parameter of that choice, the contract which exposes this choice and builds up a command
+getCommandWithCid : forall t c r t2. (Template t, Choice t c r, Template t2) => (ContractId t2 -> c) -> ContractId t2 -> (ContractId t, t) -> (AnyContractId, Command)
```
lucianojoublanc-da, line 173: Do you need all these type constraints on here? e.g. `Choice` constraint? I recall we removed it from a similar function.

lucianojoublanc-da (approved):

triggers/daml/AutoApproval.daml
brian-weir: Only when I remove the entire type signature for the function, it works. 

Removing the `choice` constraint causes `Could not deduce (HasExercise t c r0) arising from a use of ‘exerciseCmd’` error. Also, removing the type constraint while leaving the rest of the signature causes an error.

triggers/daml/AutoApproval.daml
lucianojoublanc-da: What I mean is, can you just remove `forall t c r t2. (Template t, Choice t c r, Template t2) =>`, or parts of it? Do you really need that type constraint?

brian-weir: The best I can reduce it the type constraint to without an error is `forall t c r. Choice t c r =>`.

brian-weir: I've updated this signature and a few others that could also be simplified.

PR #289, opened 2021-04-26
get quick setup to build

alexmatson-da:
This adds a regulator role template, which was necessary in order to be automated in the auto-approve trigger. Now the entire flow of requesting a regulator service, acceptance, then requesting identity verification, and acceptance is covered end-to-end by the trigger. 

I had to make some changes to Quick Setup due to the new codegen, and I got to build successfully but the identity requests made through there have broken again... I will keep looking into that later today/tmrrw

### DIFF
--- a/ui2/src/pages/login/QuickSetup.tsx
+++ b/ui2/src/pages/login/QuickSetup.tsx
@@ -46,11 +46,16 @@ import {
 } from '@daml.js/da-marketplace/lib/Marketplace/Trading/Role';
 import {
   Offer as MatchingOffer,
-  Service as MarchingService,
+  Service as MatchingService,
 } from '@daml.js/da-marketplace/lib/Marketplace/Trading/Matching/Service';
 
 import {
   Offer as RegulatorOffer,
+  Role as RegulatorRole,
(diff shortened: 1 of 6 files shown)


ui2/src/pages/login/QuickSetup.tsx
```diff
@@ -734,6 +758,8 @@ const CreateRoleContract = (props: {
             return acceptAllOffers(settlementOffers.contracts, SettlementOffer.Accept);
           case ServiceKind.LISTING:
             return acceptAllOffers(distributorOffers.contracts, DistributorOffer.Accept);
+          case ServiceKind.REGULATOR:
+            return acceptAllOffers(regulatorOffers.contracts, RegulatorOffer.Accept);
```
Sarahb-da, line 762: is this not handled in the trigger?

alexmatson-da: It does, but the quick setup page is currently working independently of the trigger afaict. The other offer accepts above are also handled by the trigger, if it is running

Sarahb-da (approved):

PR #292, opened 2021-04-27
add pooled websocket connections

alexmatson-da:
pulls in the code from the old marketplace that pools query streams into a single websocket connection. Template names are subscribed to, and the connection is reestablished with the initial message resent to include the new template ID.


alexgraham-da:
might want to use the `websockets.ts` and `queryStream.ts` from `master`, this looks like it was from before the reconnection issue was fixed and added the connection active flag.

alexmatson-da:
> might want to use the `websockets.ts` and `queryStream.ts` from `master`, this looks like it was from before the reconnection issue was fixed and added the connection active flag.

good catch, I copied the versions of those files present in the `ui` directory on `sales-eng`, instead of master. I will update that.

Sarahb-da (approved):

alexgraham-da (approved):

PR #293, opened 2021-04-27
Sales eng cleared trades

alexgraham-da:
add cleared trading to the model and the ui


ui2/src/pages/network/Clearing.tsx
```diff
@@ -23,13 +28,23 @@ export const ClearingServiceTable: React.FC<Props> = ({ services }) => {
...
+    MarketService
+  );
 
   const roles = useStreamQueries(Role).contracts;
   const hasRole = roles.length > 0 && roles[0].payload.provider === party;
 
-  const terminateService = async (c: CreateEvent<Service>) => {
-    await ledger.exercise(Service.Terminate, c.contractId, { ctrl: party });
+  const terminateService = async (c: CreateEvent<Service> | CreateEvent<MarketService>) => {
+    if (getTemplateId(c.templateId) === 'Marketplace.Clearing.Service.Service') {
```
alexgraham-da, line 43: does anyone know a better way to check/coerce the type here? I couldn't quite figure out how to do it using `typeof` with the union types having type parameters, but it felt like it should be possible.

daml/Marketplace/Listing/Model.daml
```diff
@@ -10,11 +11,17 @@ data Status
     | Disabled
   deriving (Eq, Show)
 
+data ListingType
+    = Cleared { clearinghouse: Party, approvalCid : (ContractId ClearedListingApproval) }
```
brian-weir, line 15:
```suggestion
    = Cleared with 
         clearinghouse: Party
         approvalCid: ContractId ClearedListingApproval
```

daml/Marketplace/Trading/Model.daml
```diff
@@ -63,6 +63,12 @@ data Execution = Execution with
 
 type T = Order
 
+data TradeCollateral = Cleared { clearinghouse: Party } | Collateral (ContractId AssetDeposit) deriving (Show, Eq)
```
brian-weir, line 66:
```suggestion
data TradeCollateral 
  = Cleared with clearinghouse: Party 
  | Collateral (ContractId AssetDeposit) 
    deriving (Show, Eq)
```

daml/Marketplace/Trading/Confirmation/Model.daml
```diff
@@ -47,6 +48,28 @@ template Confirmation
...
+              isPartialFill = execution.quantity < order.remainingQuantity
+              executions = execution :: order.executions
+              remainingQuantity = order.remainingQuantity - execution.quantity
+              quantities = [case order.details.side of
+                Order.Buy  -> execution.quantity * execution.price
+                Order.Sell -> execution.quantity]
+
+            if isPartialFill
+              then do
+                create order with executions, remainingQuantity, collateral = order.collateral, status = Order.PartiallyExecuted
```
brian-weir, line 68: Do you need to set `collateral = order.collateral` since its already on the order?

```suggestion
                create order with executions, remainingQuantity, status = Order.PartiallyExecuted
```

ui2/src/pages/clearing/Members.tsx
```diff
@@ -115,6 +131,51 @@ const ClearingMembersComponent: React.FC<RouteComponentProps & ServicePageProps<
           };
         })}
       />
+      <Header as="h2">Cleared Trades</Header>
+      <StripedTable
+        loading={clearedTradesLoading}
+        rowsClickable
```
Sarahb-da, line 137: I don't think `rowsClickable` needs to be here - unless u provide an `onClick` function (plus, there is a button in the row data)

ui2/src/pages/clearing/Members.tsx
```diff
@@ -115,6 +131,51 @@ const ClearingMembersComponent: React.FC<RouteComponentProps & ServicePageProps<
...
+                Novate Trade
+              </Button>,
+            ],
+          };
+        })}
+      />
+      <Header as="h2">Cleared Trade Sides</Header>
+      <StripedTable
+        loading={clearedTradeSidesLoading}
+        rowsClickable
```
Sarahb-da, line 163: ex: here `rowsClickable` works because there are no buttons in the row data, and the onClick function will apply to the whole row

ui2/src/pages/listing/New.tsx
```diff
@@ -8,14 +8,18 @@ import { AssetDescription } from '@daml.js/da-marketplace/lib/Marketplace/Issuan
...
-import { ServicePageProps } from '../common';
+import { Service as MarketClearingService } from '@daml.js/da-marketplace/lib/Marketplace/Clearing/Market/Service';
+import { publicParty, usePartyName } from '../../config';
+import { ServicePageProps, createDropdownProp } from '../common';
 import { Button, Form, Header, Icon } from 'semantic-ui-react';
 import FormErrorHandled from '../../components/Form/FormErrorHandled';
 import { IconClose } from '../../icons/icons';
 import Tile from '../../components/Tile/Tile';
 
+const COLLATERALIZED_VALUE = 'COLLATERALIED_MARKET';
```
Sarahb-da, line 21: `COLLATERALIZED`

ui2/src/pages/listing/New.tsx
```diff
@@ -172,9 +184,21 @@ const NewComponent: React.FC<RouteComponentProps & ServicePageProps<Service>> =
...
+            options={[
+              createDropdownProp('-- Collateralized Market --', COLLATERALIZED_VALUE),
+              ...clearedMarketServices.map(cms => createDropdownProp(getName(cms.payload.provider), cms.payload.provider)),
+            ]}
+            onChange={(_, change) => setClearedBy(change.value as string)}
+          />
           <Form.Input label="Trading Calendar ID" required readOnly placeholder={calendarId} />
           <div className="submit">
-            <Button type="submit" className="ghost" disabled={!canRequest} content="Submit" />
+            <Button type="submit" className="ghost" disabled={false} content="Submit" />
```
Sarahb-da, line 201: possibly left over from debugging - if not no need for `disabled={false} ` (afaik)

ui2/src/pages/network/Clearing.tsx
Sarahb-da: what do u want to check the type of?

Sarahb-da: p.s. it might be nice to have all the right hand side template name strings pulled out as variables or put in an enum or something

daml/Marketplace/Trading/Model.daml
```diff
@@ -73,9 +79,55 @@ template Order
...
+
+          takerCid <- create ClearedTradeSide with
+            exchange = provider
+            order = takerOrder
+            participant = takerOrder.customer
+            ..
+
+          return (makerCid, takerCid)
+
+template ClearedTradeSide
```
brian-weir, line 123: Whats the purpose of this `ClearedTradeSide` ?

daml/Marketplace/Trading/Service.daml
```diff
@@ -25,8 +31,46 @@ template Service
     key (operator, provider, customer) : (Party, Party, Party)
     maintainer key._1
 
+    let
+      withdrawFromAllocationAccount : Order.TradeCollateral -> Update Order.TradeCollateral
+      withdrawFromAllocationAccount = \tc -> case tc of
```
brian-weir, line 36:
```suggestion
      withdrawFromAllocationAccount = \case
```

ui2/src/pages/network/Clearing.tsx
```diff
@@ -169,58 +203,84 @@ export const ClearingServiceTable: React.FC<Props> = ({ services }) => {
...
+                </Button.Group>,
+              ],
+            };
+          }),
+          ...marketOffers.map((c, i) => {
+            return {
+              elements: [
+                getTemplateId(c.templateId),
+                getName(c.payload.customer),
+                <Button.Group>
```
Sarahb-da, line 264: do we need a button group?

Sarahb-da: (personally I like the semantic-ui buttons better, which are just 
```
 <Button
          className="ghost"
          onClick={() => acceptOffer(c)}
/>
```

ui2/src/pages/network/Clearing.tsx
alexgraham-da: I have the type of `c` as `CreateEvent<Service> | CreateEvent<MarketService>` so I can use the same function when mixing the two types of offers in the same table here (since they have the same fields and choice structure):  https://github.com/digital-asset/da-marketplace/blob/04b517d2970214ae7f0e33f2c86d5c4bf1791021/ui2/src/pages/network/Clearing.tsx#L160.

So it would be nice to switch on the type of `c` without having to check the equality of the template name as a string.

Sarahb-da (commented):
A few very minor comments. Nice work!

daml/Marketplace/Trading/Confirmation/Model.daml
alexgraham-da: ah yes, good catch

ui2/src/pages/listing/New.tsx
alexgraham-da: good catch, thank you

ui2/src/pages/listing/New.tsx
alexgraham-da: spelling is hard, but I guess this is why consts instead of just directly comparing strings is good...

daml/Marketplace/Trading/Model.daml
alexgraham-da: This is so the clearinghouse and the participant has a record that the trade was novated and is the representation of the contract between the participant and the clearinghouse (as central counterparty) for this particular trade. In practice this is what baymarkets is using to keep track of trades to perform mtm/margin adjustments. This was ported over from the old model, so I'm open to different ways of handling this.

ui2/src/pages/network/Clearing.tsx
alexgraham-da: These are now semantic buttons but some of the props may be left over from the material versions. I guess it doesn't need the `Button.Group`, I just had that since some of the other tables on this page have multiple buttons, I'll try without!

daml/Marketplace/Trading/Model.daml
brian-weir: Would the performing of mtm/margin adjustments be performed by another service ?

alexgraham-da: Hmmm...the actual transfers related to the adjustments are done by `Clearing.Service`, are you suggesting a separate service might be responsible for doing the calculations?

daml/Marketplace/Trading/Service.daml
```diff
@@ -25,8 +31,46 @@ template Service
...
+            providerOrderId = None
+            executions = []
+            remainingQuantity = details.asset.quantity
+            collateral = Order.Cleared clearinghouse
+
+          optClearingService <- lookupByKey @Clearing.Service (operator,clearinghouse,customer)
+
+          case optClearingService of
+            None -> do
+              let reason = Error with code = "1234"; message = "Customer is not a member of clearinghouse"
```
alexmatson-da, line 60: Will we want to create different error codes for different errors?

alexgraham-da: I think the initial reason for having error codes comes from Exberry returning error codes - so it could be good to do keep track of their codes somehow, and add our own. Looks like all of exberry's codes are in the 1000s so the marketplace model specific ones could be in the 3000s or something to that effect

ui2/src/pages/network/Clearing.tsx
alexgraham-da: > p.s. it might be nice to have all the right hand side template name strings pulled out as variables or put in an enum or something (this one and the one on line 69 and 82 etc.)

Yup that makes sense, it's too bad the codegen doesn't give anything like that.

daml/Marketplace/Trading/Model.daml
brian-weir: I was just wondering who was acting on this template. Maybe it would be clearer to put it in a `Clearing.Model` and call it `ClearingInstruction` or something like that ?

lucianojoublanc-da: I also think that this is leaking Baymarkets specifics into this module. 

It's mixing trading concerns with clearing concerns. I think you should endeavour to keep them separate.

Could the concept of novation be made more general?

daml/Marketplace/Trading/Model.daml
lucianojoublanc-da: Are these really mutually exclusive? Couldn't you have a cleared trade that requires collateral also?

daml/Marketplace/Trading/Service.daml
```diff
@@ -25,8 +31,46 @@ template Service
...
 
+    let
+      withdrawFromAllocationAccount : Order.TradeCollateral -> Update Order.TradeCollateral
+      withdrawFromAllocationAccount = \tc -> case tc of
+        (Order.Cleared party)         -> return $ Order.Cleared party
+        (Order.Collateral depositCid) -> Order.Collateral <$> exerciseByKey @AllocationAccountRule.T allocationAccount.id AllocationAccountRule.Withdraw with transferTo = tradingAccount, ..
+
     controller customer can
-      nonconsuming RequestCreateOrder : (ContractId Order.T, ContractId CreateOrderRequest, ContractId AssetDeposit)
+      nonconsuming RequestCreateClearedOrder : Either (ContractId Order.T)
```
lucianojoublanc-da, line 41: Doesn't feel right to me that we have a second function for creating cleared order. As per previous comment, I think we need to separate the concept of an order from collateral/clearing. Then have a single `RequestCreateOrder` with collateral/clearing as additional arguments, separate from the order economics.

lucianojoublanc-da (commented):
I would think about how to separate the clearing and settlement concerns such as margin and collateral from the order and trade economics.

I think we're leaking the Baymarkets model with `ClearedTradeSide`. As you suggested, maybe moving this stuff into a separate service would make sense.

And from a UI point of view, I don't think you want to have different functions for submitting cleared/uncleared orders.

daml/Marketplace/Trading/Service.daml
alexgraham-da: Makes sense - I originally had it that way and split it out to be a bit cleaner before I had the `TradingCollateral` type, so I can put it back into one function.

daml/Marketplace/Trading/Model.daml
alexgraham-da: Just to clarify, `ClearedTradeSide` was a v1 marketplace concept that Baymarkets was using, not necessarily a Baymarkets specific data structure that we put in for them. I do agree though that it may make more sense renamed in the clearing module, since at this point it isn't something the trading service cares about or needs to keep track of.

daml/Marketplace/Trading/Model.daml
alexgraham-da: Hmmm not 100% sure. I figured with a cleared trade in this context, the "collateral" is the membership in the clearinghouse and whatever accounts you hold with them. Maybe `Collateral` vs `Noncollateralized` might make sense? 

In that case, instead of passing the clearinghouse around with the order, the order choice could pull up the listing and check if a clearing membership/approval is necessary instead, regardless of if a collateral is used or not.

daml/Marketplace/Trading/Model.daml
alexgraham-da: Novation could definitely be more generalized, although I think it's out of the scope of this PR.

alexgraham-da: To the comment: "I would think about how to separate the clearing and settlement concerns such as margin and collateral from the order and trade economics." this may need a larger restructuring since the collateral as a `ContractId AssetDeposit` was already integrated into the `Trading` service. Something else that may make sense to refactor in another PR.

daml/Marketplace/Trading/Model.daml
alexgraham-da: Actually, looking up the listing in the order creation choice is going to require multi-party submits in daml hub since it will require `readAs: [public]`, so I don't think this will be possible yet.

ui2/src/pages/network/Clearing.tsx
Sarahb-da: ohh I see what you are saying now

Sarahb-da: ohh I see what you are saying now. Possibly `instanceof` ? Im doing some googling..

daml/Marketplace/Trading/Model.daml
alexgraham-da: @brian-weir @lucianojoublanc-da - I think a good solution to this, as well as for better separation of the clearing/collateral concerns can be to rely on the `Listing` for making decisions such as: what collateral is necessary, if a clearinghouse needs to approve the trade, how the trade gets settled, etc. Which makes sense to me, since the listing is where all of that should be specified anyway.

So in this case: `TradeCollateral` could just be something like `optCollateral: Optional (ContractId AssetDeposit)` (or still a special type for readability), and the order request choice can lookup the listing information and see if a collateral is required and reject the order if it does not exist. If the listing requires the order to be cleared by a clearinghouse, that can happen there as well. This would also allow for an order to require collateral _and_ require clearing. Then once the orders are matched and get to the matching service, the matching service can get the listing information as well and process the trade accordingly.

If that seems like a good solution, then the question for implementing it is: should the `Order` have the `ContractId Listing`, the `Listing`, or just the relevant info? If it's the contract id or key info, that won't currently work on hub since it would require `readAs: [publicParty]` since the `customer` isn't a direct observer on a `Listing`. Let me know what you think, I'm open to other solutions as well.

lucianojoublanc-da: I do like the idea of linking this with the listing somehow. It makes sense from a business point of view, as I think it's where this would be determined in the real world. So you could originate the same instrument and then have it traded multiple ways depending on the listing.

However this will prevent e.g. a market maker from clearing the trade vs. a user trading fully collateralised on the same order-book; you'll have two separate instruments. I think this scenario was an original business requirement.

> that won't currently work on hub since it would require readAs: [publicParty] since the customer isn't a direct observer on a Listing

So how does it currently have the asset details in there? I assume the `Asset` is just embeded in `Order`?
Yeah, that does seem like a problem.

This sounds like a considerable refactor. Is it worth doing this separately, so we don't hold up this merge? I do definitely think that it's worth doing it though.

brian-weir: @alexgraham-da - I think your idea is sound. Ideally, the `Trading` service shouldn't care if the order is traded on margin, cleared, fully collateralised, etc. It's a much cleaner separation if we move these details onto the `Listing` contract and have these `Pre-Trade` checks against an order request live on the `Listing` service. There are some existing TODOs around checking precision, max/min quantity and even the status of the listing against an order.

> So how does it currently have the asset details in there? I assume the Asset is just embeded in Order?

Currently, an order is linked to a listing via the listings 'symbol' (which is also part of the contract key on `Listing` as `listingId` which allows for lookups). Sending the same symbol twice to Exberry will cause an error. 

> However this will prevent e.g. a market maker from clearing the trade vs. a user trading fully collateralised on the same order-book; you'll have two separate instruments. I think this scenario was an original business requirement.

To satisfy this requirement, we'll need a way of identifying market-makers from other users who must put up full collateral for a trade. There's multiple ways of implementing this requirement but I don't think we should have multiple listings/instruments. I think it'll require a bit more design and should be part of another PR. I've added this point to the spreadsheet so we have it tracked there for now.

alexgraham-da: > Currently, an order is linked to a listing via the listings 'symbol' (which is also part of the contract key on Listing as listingId which allows for lookups).

I was looking at this, and I don't think this will work 100% for looking up a `Listing` in all places - the key of the `Listing` is `(operator, provider, listingId)`. Since the provider of the `Trading` service isn't necessarily the provider of the `Listing`, you wouldn't have a way of knowing which `provider` to use for the key. Maybe it would make more sense for a `Listing` to be keyed by an `Id`? That way you can store the `Id` in contracts like `Order` that need to do a lookup.

alexgraham-da: > Ideally, the Trading service shouldn't care if the order is traded on margin, cleared, fully collateralised, etc. It's a much cleaner separation if we move these details onto the Listing contract and have these Pre-Trade checks against an order request live on the Listing service. There are some existing TODOs around checking precision, max/min quantity and even the status of the listing against an order.

I think then this is indeed the way to go. It's not a _huge_ refactor to do just this, but I think it makes the most sense to get this PR out now and then refactor and add this functionality separately (especially since we'll need to have a way of handling visibility issues with the `Listing` on Hub).

The market maker vs. regular user requirement is an interesting one - I'd say that's definitely separate and we can decide later if it would need to be rolled into the listing pre-trade functionality at the same time, or just design that in a way that would eventually allow it.

Sarahb-da (approved):
lgtm!

Issue #291, opened 2021-04-27
lifecycle payout will fail when settling multiple cashflows


lucianojoublanc-da:
> gabor.hosszu  2:25 PM
> ```
>           payoutCids <- forA pending \(qty, asset) -> do
>             payoutDepositCid :: _ <- exercise safekeepingDepositCid $ AssetDeposit_Split [qty * assetDeposit.asset.quantity]           <<<<--------
>             exerciseByKey @AssetSettlementRule assetDeposit.account.id $
>                AssetSettlement_AddController description.issuer
>             cid <- exerciseByKey @AssetSettlementRule description.safekeepingAccountId (
>               AssetSettlement_Transfer with
>                 receiverAccountId = assetDeposit.account.id
>                 depositCid = payoutDepositCid)
>             exerciseByKey @AssetSettlementRule assetDeposit.account.id $
>               AssetSettlement_RemoveController description.issuer
>             pure cid
> ```
> is it not a problem when there are multiple assets in pending and one tries to split `safekeepingDepositCid ` more than once?
> 2:29
> so that should be rather a fold
> 
> Luciano  2:37 PM
> You mean that after the first split, the original `safekeepingDepositCid` will have been archived? I think you're right. Good spot.

Thanks for the spot @gaborh-da. As discussed we were thinking of changing this from doing the transfer directly, to instead returning a settlement instruction that is then cleared as a separate step.

lucianojoublanc-da:
Fixed by https://github.com/gaborh-da/da-marketplace/pull/2

Issue #290, opened 2021-04-27
lifecycle does not create new `AssetDescription` records


lucianojoublanc-da:
It appears that the lifecycle function

https://github.com/digital-asset/da-marketplace/blob/sales-eng/daml/Marketplace/Custody/Model.daml#L115

Is not creating subsequent `AssetDescription` versions.

To recap, the process should be:

  1. Create a single `AssetDescription` with `assetId.version == 0` on origination.
  2. On lifecycle, add a new entry into `Index` to allow a reverse lookup.
  3. Also create a new `AssetDescription` with the post-lifecycle version, if it doesn't exist already.

It seems that step 3 is being skipped.

~~I'm considering a refactor as follows:~~

```
template Index
  with
    assetLabel: Text
    descriptionSignatories: Set Party
    version: Version
    claims: Claims
  where
    signatory descriptionSignatories
    key (descriptionSignatories, assetLabel, claims) : (Set Party, Text, Claims)
    maintainer key._1
```
~~This would remove the single monolithic `Map` and would likely improve performance. The record can then be looked up with the standard `lookupByKey`, and if it's `None`, then a new `Index`, as well as the missing post-lifecycle `AssetDescription` could be created.~~

cc @gaborh-da for discovering!

lucianojoublanc-da:
We could also consider referencing `Index` in `AssetDescription`, so we don't need to keep duplicates of the `Claim` tree.

gaborh-da:
I tried with this:
```
module Tests.Bond where
-- ^ This module tests the full lifecycle of a binary option

import Common (Customer, Providers, onboardProviders, onboardCustomer)
import ContingentClaims.Claim (serialize)
import DA.Date (date, Month(Jan, Oct))
import DA.Finance.Types (Id(..), Asset(..))
import DA.Finance.Asset (AssetDeposit)
import DA.Finance.Asset.Settlement (AssetSettlementRule(..), AssetSettlement_AddController(..), AssetSettlement_RemoveController(..), AssetSettlement_Transfer(..))
import Daml.Script
import DA.Next.Set qualified as Set
import DA.Map qualified as Map
import Marketplace.Issuance.Service (RequestOrigination(..), Originate(..), RequestCreateIssuance(..), CreateIssuance(..))
import Marketplace.Custody.Service (Service, RequestTransferDeposit(..), TransferDeposit(..), RequestLifecycle(..), Lifecycle(..), RequestCreditAccount(..), CreditAccount(..))
[...]

PR #295, opened 2021-04-28
Failing asset descriptor key

gaborh-da:


daml/Marketplace/Custody/Model.daml
```diff
@@ -113,9 +114,13 @@ template LifecycleRequest
...
               election _ _ = deserialize choice -- will throw if the claims don't match
           Lifecycle.Result{remaining, pending} <- Lifecycle.lifecycle spot election (deserialize description.claims) t
-          (_, postAssetVersion) <- exercise versionsCid . AssetDescription.LookupOrInsert $ serialize remaining
-          payoutCids <- forA pending \(qty, asset) -> do
-            payoutDepositCid :: _ <- exercise safekeepingDepositCid $ AssetDeposit_Split [qty * assetDeposit.asset.quantity]
+          (_, postAssetVersion, _) <- exercise versionsCid AssetDescription.LookupOrInsert with newClaims = serialize remaining, assetDescription = description
+          let payoutSplitQuantities = (\(qty, _) -> qty * assetDeposit.asset.quantity) <$> pending
+          -- Next step could be:
+          -- create SettlementInstruction with buyer = customer; seller = issuer; ..
+          splitResultDeposits <- exercise safekeepingDepositCid AssetDeposit_Split with quantities = payoutSplitQuantities
```
lucianojoublanc-da, line 121: So if I understand correctly, you're letting `AssetDeposit_Split` handle the splitting by passing it a list instead of calling it individually each time, correct?

daml/Marketplace/Custody/Model.daml
```diff
@@ -113,9 +114,13 @@ template LifecycleRequest
...
+          let payoutCids = init splitResultDeposits
+          payoutCids <- forA payoutCids \payoutDepositCid -> do
```
lucianojoublanc-da, line 123: Just my opinion, but I think the de-structuring is more legible than using `init` here. (Plus you save an import :wink: )

daml/Marketplace/Custody/Model.daml
lucianojoublanc-da, line 121:
```suggestion
          splitResultDeposits <- exercise safekeepingDepositCid $ AssetDeposit_Split payoutSplitQuantities
```

daml/Marketplace/Issuance/AssetDescription.daml
```diff
@@ -55,11 +55,16 @@ template Index
     maintainer key._1
 
     controller descriptionSignatories can
-      LookupOrInsert : (ContractId Index, Version) with claims : Claims
+      LookupOrInsert : (ContractId Index, Version, Optional (ContractId AssetDescription)) with
```
lucianojoublanc-da, line 58: ~~I don't understand why you're changing this, if you're not using the return value, up on line 117.~~ Ok so this is to create the post-lifecycle `AssetDescription` version?
I don't think it's a good idea to hide it in here; it becomes a side-effect you need to know as a programmer.

Also beware that we may change this so that we have one `Index` per `AssetLabel` / `Version`, instead of having them all in a  single `Map`.

lucianojoublanc-da (commented):
I suggest that for the time being you don't update `LookupOrInsert`, as we may need to remove that altogether (see https://github.com/digital-asset/da-marketplace/issues/290 ). Just do the `AssetDescription` create separately.

daml/Marketplace/Custody/Model.daml
gaborh-da: Exactly

daml/Marketplace/Custody/Model.daml
gaborh-da: May I ask how exactly you mean de-structuring? I am probably missing something but in this case 'we need everything but the last element' from the list.

daml/Marketplace/Issuance/AssetDescription.daml
gaborh-da: > Ok so this is to create the post-lifecycle AssetDescription version?
Yes.

> I don't understand why you're changing this, if you're not using the return value, up on line 117. 
Yes, probably the return value is not even needed so I could just change back.

> I don't think it's a good idea to hide it in here; it becomes a side-effect you need to know as a programmer.
That's true, where would you do it, in Lifecycle/Process?

daml/Marketplace/Custody/Model.daml
gaborh-da: If you insist on changing that, I can do it but I deliberately changed it, in my opinion it is more readable (despite being a bit longer but being more explicit and DAML-ish)

daml/Marketplace/Custody/Model.daml
lucianojoublanc-da: So if you look at the diff, this is replacing
```
payoutDepositCid :: _ <- exercise safekeepingDepositCid $ AssetDeposit_Split [qty * assetDeposit.asset.quantity]
```
The right-hand-side of the arrow is taking the head of the tail for you. In this case `_` is the result of `init`.

lucianojoublanc-da: Oh sorry, my bad here, I thought you were taking the tail. Pls ignore this comment.

daml/Marketplace/Issuance/AssetDescription.daml
lucianojoublanc-da: Probably in Process. But see my general comment about #291

daml/Marketplace/Custody/Model.daml
```diff
@@ -113,9 +115,17 @@ template LifecycleRequest
               election : Claim Observation Date Id -> Claim Observation Date Id -> Claim Observation Date Id
               election _ _ = deserialize choice -- will throw if the claims don't match
           Lifecycle.Result{remaining, pending} <- Lifecycle.lifecycle spot election (deserialize description.claims) t
-          (_, postAssetVersion) <- exercise versionsCid . AssetDescription.LookupOrInsert $ serialize remaining
-          payoutCids <- forA pending \(qty, asset) -> do
-            payoutDepositCid :: _ <- exercise safekeepingDepositCid $ AssetDeposit_Split [qty * assetDeposit.asset.quantity]
+          let serializedClaims = serialize remaining
```
lucianojoublanc-da, line 118: Small detail, but I'd rename this something like `remaining'` just to make it clear it's the remaining claims, not the original ones.

gaborh-da: Done.

daml/Marketplace/Issuance/AssetDescription.daml
gaborh-da: See now.

gaborh-da:
@lucianojoublanc-da is there anything else to do with this PR?

gaborh-da:
Next ones:
https://github.com/gaborh-da/da-marketplace/pull/1
https://github.com/gaborh-da/da-marketplace/pull/2

daml/Tests/Bond.daml
```diff
@@ -0,0 +1,115 @@
...
+
+  [(_, bobsOption)] <- queryFilter @AssetDeposit bob.customer (\c -> c.asset.id.label == "BondLabel")
+  2 === bobsOption.asset.id.version
+  bobsMoneyWithCids <- getMoneyAssetsFor bob.customer
+  couponPayout + notional + couponPayout ===
+      sum (map (\(_, money) -> money.asset.quantity) bobsMoneyWithCids)
+  [(_, alicesMoney)] <- getMoneyAssetsFor alice
+  usdAmount - couponPayout - notional - couponPayout === alicesMoney.asset.quantity
+
+getMoneyAssetsFor party = do
```
alexgraham-da, line 67: `do` is redundant here, and it would be nice to have a type signature for this function

daml/Tests/Bond.daml
```diff
@@ -0,0 +1,115 @@
...
+  lifecycleReqCid <- submit bob.customer $ exerciseCmd bob.custodyServiceCid RequestLifecycle with
+    assetDepositCid = bondCid
+    choice = bobsDesc.claims
+  bondCid :: _ <- submit providers.bank $ exerciseCmd custodyServiceCid $ Lifecycle with
+    safekeepingDepositCid = corpActDepositCid
+    lifecycleRequestCid = lifecycleReqCid
+    fixings = Map.fromList []
+
+  [(_, bobsOption)] <- queryFilter @AssetDeposit bob.customer (\c -> c.asset.id.label == "BondLabel")
+  1 === bobsOption.asset.id.version
```
alexgraham-da, line 43: might be personal preference, but elsewhere in `Tests` these types of asserts are `varToCheck === desiredValue` (which is more readable imo)

daml/Tests/Bond.daml
```diff
@@ -0,0 +1,115 @@
...
+  (optionDescCid, optionDesc) <- submit providers.bank $ exerciseCmd alice.issuanceServiceCid $ Originate origReqCid
+
+  -- Alice issues the bond
+  createReqCid <- submit alice.customer $ exerciseCmd alice.issuanceServiceCid
+    RequestCreateIssuance with
+      issuanceId = "Myissuance"
+      accountId = Id (Set.fromList [alice.customer, providers.bank]) alice.mainAccount.id.label 0
+      assetId = optionDesc.assetId
+      quantity = 1.0
+  (_issuanceCid, bondCid) <- submit providers.bank $ exerciseCmd alice.issuanceServiceCid $ CreateIssuance createReqCid
```
alexgraham-da, line 100: was this supposed to be `(_, bondCid)`?

alexgraham-da (approved):
A few small comments, but otherwise looks good to me. Thanks for doing this!

alexmatson-da (approved):
thanks!

daml/Tests/Bond.daml
gaborh-da: done

daml/Tests/Bond.daml
gaborh-da: I was thinking like
```
assertEquals(double expected, double actual)
```
(JUnit)

But reversed.

daml/Tests/Bond.daml
gaborh-da: yes, done

gaborh-da:
@alexgraham-da I think I managed to fix the commented pieces but I cannot merge. Could you please do it (if no more changes needed)?

alexgraham-da:
> @alexgraham-da I think I managed to fix the commented pieces but I cannot merge. Could you please do it (if no more changes needed)?

Done!

PR #300, opened 2021-05-03
Sales eng add fair value

alexgraham-da:
add the ability for the provider/customer of the `Clearing.Market.Service` to request `FairValue` contracts for listings

A `FairValue` is calculated by:
1. The price of the most recent trade before the specified 'upTo'
2. If there are no trades, the most recent 'FairValue'
3. If there are no 'FairValue', request manual calculation

### DIFF
--- a/daml/Marketplace/Clearing/Market/Model.daml
+++ b/daml/Marketplace/Clearing/Market/Model.daml
@@ -12,13 +12,14 @@ template FairValueCalculationRequest
     currency : Id
     upTo : Time
   where
-    signatory provider
+    signatory operator, provider
     controller provider can
       FairValueCalculationRequest_Ack : ()
         do return ()
 
 template ManualFairValueCalculation
   with
+    operator : Party
     provider : Party
     customer : Party
     listingId : Text
@@ -26,7 +27,7 @@ template ManualFairValueCalculation
     upTo : Time
     observers : Set Party
   where
-    signatory provider
+    signatory operator, provider
 
     controller customer can
       ManualFairValueCalculation_Calculate : ContractId FairValue
@@ -38,6 +39,7 @@ template ManualFairValueCalculation
 
 template FairValue
   with
+    operator : Party
     provider : Party
     customer : Party
     listingId : Text
@@ -47,5 +49,5 @@ template FairValue
     upTo : Time
     observers : Set Party
   where
-    signatory provider
+    signatory operator, provider
     observer customer, observers
--- a/daml/Marketplace/Clearing/Market/Model.daml
+++ b/daml/Marketplace/Clearing/Market/Model.daml
@@ -1,0 +1,51 @@
+module Marketplace.Clearing.Market.Model where
+
+import DA.Finance.Types
+import DA.Next.Set (Set)
+
+template FairValueCalculationRequest
+  with
+    operator : Party
+    provider : Party
+    customer : Party
+    listingId : Text
+    currency : Id
+    upTo : Time
+  where
+    signatory provider
+    controller provider can
+      FairValueCalculationRequest_Ack : ()
+        do return ()
+
+template ManualFairValueCalculation
+  with
+    provider : Party
+    customer : Party
+    listingId : Text
+    currency : Id
+    upTo : Time
+    observers : Set Party
+  where
+    signatory provider
+
+    controller customer can
+      ManualFairValueCalculation_Calculate : ContractId FairValue
+        with
+          price : Decimal
+        do
+          timestamp <- getTime
+          create FairValue with ..
+
+template FairValue
+  with
+    provider : Party
+    customer : Party
+    listingId : Text
+    price : Decimal
+    currency : Id
+    timestamp : Time
+    upTo : Time
+    observers : Set Party
+  where
+    signatory provider
+    observer observers
--- a/daml/Marketplace/Clearing/Market/Model.daml
+++ b/daml/Marketplace/Clearing/Market/Model.daml
@@ -28,8 +28,9 @@ template ManualFairValueCalculation
     observers : Set Party
   where
     signatory operator, provider
+    let exchange = customer
 
-    controller customer can
+    controller exchange can
       ManualFairValueCalculation_Calculate : ContractId FairValue
         with
           price : Decimal
--- a/daml/Marketplace/Clearing/Market/Model.daml
+++ b/daml/Marketplace/Clearing/Market/Model.daml
@@ -8,7 +8,7 @@ template FairValueCalculationRequest
     operator : Party
     provider : Party
     customer : Party
-    listingId : Text
+    optListingIds : Optional [Text]
     currency : Id
     upTo : Time
   where
@@ -48,4 +48,4 @@ template FairValue
     observers : Set Party
   where
     signatory provider
-    observer observers
+    observer customer, observers
--- a/daml/Marketplace/Clearing/Market/Service.daml
+++ b/daml/Marketplace/Clearing/Market/Service.daml
@@ -23,7 +23,7 @@ template Service
     nonconsuming choice RequestFairValues : ContractId FairValueCalculationRequest
       with
         party : Party
-        listingId : Text
+        optListingIds : Optional [Text]
         currency : Id
       controller party
       do
--- a/daml/Marketplace/Clearing/Market/Service.daml
+++ b/daml/Marketplace/Clearing/Market/Service.daml
@@ -1,9 +1,13 @@
 module Marketplace.Clearing.Market.Service where
 
+import DA.Finance.Types
+
+import Marketplace.Clearing.Market.Model
 import Marketplace.Listing.Model
 import Marketplace.Utils
 
 import DA.Next.Set (Set)
+import DA.Next.Set qualified as Set
 
 template Service
   with
@@ -16,8 +20,18 @@ template Service
     key (operator, provider, customer) : (Party, Party, Party)
     maintainer key._1
 
-    controller customer can
+    nonconsuming choice RequestFairValues : ContractId FairValueCalculationRequest
+      with
+        party : Party
+        listingId : Text
+        currency : Id
+      controller party
+      do
+        upTo <- getTime
+        assert $ Set.member party $ Set.fromList [provider, customer]
+        create FairValueCalculationRequest with ..
 
+    controller customer can
       nonconsuming ApproveClearedListing : ContractId ClearedListingApproval
         with
           listingId : Text
--- a/daml/Marketplace/Clearing/Market/Service.daml
+++ b/daml/Marketplace/Clearing/Market/Service.daml
@@ -20,16 +20,26 @@ template Service
     key (operator, provider, customer) : (Party, Party, Party)
     maintainer key._1
 
-    nonconsuming choice RequestFairValues : ContractId FairValueCalculationRequest
+    nonconsuming choice RequestAllFairValues : ContractId FairValueCalculationRequest
       with
         party : Party
-        optListingIds : Optional [Text]
         currency : Id
       controller party
       do
         upTo <- getTime
         assert $ Set.member party $ Set.fromList [provider, customer]
-        create FairValueCalculationRequest with ..
+        create FairValueCalculationRequest with optListingIds = None; ..
+
+    nonconsuming choice RequestFairValues : ContractId FairValueCalculationRequest
+      with
+        party : Party
+        listingIds : [Text]
+        currency : Id
+      controller party
+      do
+        upTo <- getTime
+        assert $ Set.member party $ Set.fromList [provider, customer]
+        create FairValueCalculationRequest with optListingIds = Some listingIds; ..
 
     controller customer can
       nonconsuming ApproveClearedListing : ContractId ClearedListingApproval
--- a/daml/Marketplace/Clearing/Market/Service.daml
+++ b/daml/Marketplace/Clearing/Market/Service.daml
@@ -41,6 +41,27 @@ template Service
         assert $ Set.member party $ Set.fromList [provider, customer]
         create FairValueCalculationRequest with optListingIds = Some listingIds; ..
 
+    controller provider can
+      nonconsuming CreateFairValue : ContractId FairValue
+        with
+          listingId : Text
+          price : Decimal
+          currency : Id
+          timestamp : Time
+          upTo : Time
+          observers : Set Party
+        do
+          create FairValue with ..
+
+      nonconsuming CreateManualFairValueRequest : ContractId ManualFairValueCalculation
+        with
+          listingId : Text
+          currency : Id
+          upTo : Time
+          observers : Set Party
+        do
+          create ManualFairValueCalculation with ..
+
     controller customer can
       nonconsuming ApproveClearedListing : ContractId ClearedListingApproval
         with
--- a/triggers/daml/ClearingTrigger.daml
+++ b/triggers/daml/ClearingTrigger.daml
@@ -8,12 +8,17 @@ import Daml.Trigger
 import DA.Finance.Asset
 import DA.Finance.Types
 
+import qualified Marketplace.Listing.Model as Listing
+import qualified Marketplace.Clearing.Market.Model as Clearing
 import qualified Marketplace.Clearing.Model as Clearing
 import qualified Marketplace.Clearing.Role as Clearing
 import qualified Marketplace.Clearing.Service as Clearing
 import  Marketplace.Clearing.Service
 
 import Utils
+import DA.List (head)
+import DA.List.Total (sortOn)
+import DA.Time (seconds, subTime)
 
 data ClearingState = ClearingState {
    operatorParty : Party
@@ -26,7 +31,10 @@ handleClearing = Trigger
   , rule = handleCCPRule
   , registeredTemplates = RegisteredTemplates  [ registeredTemplate @Clearing.Role
                                                , registeredTemplate @Clearing.Service
+                                               , registeredTemplate @Listing.T
                                                , registeredTemplate @AssetDeposit
+                                               , registeredTemplate @Clearing.FairValueCalculationRequest
+                                               , registeredTemplate @Clearing.ClearedTradeSide
                                                , registeredTemplate @Clearing.MarginCalculation
                                                , registeredTemplate @Clearing.MarkToMarketCalculation ]
   , heartbeat = None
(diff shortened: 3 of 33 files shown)


daml/Marketplace/Clearing/Market/Model.daml
```diff
@@ -0,0 +1,51 @@
...
+template FairValueCalculationRequest
+  with
+    operator : Party
+    provider : Party
+    customer : Party
+    optListingIds : Optional [Text]
+    currency : Id
+    upTo : Time
+  where
+    signatory provider
```
brian-weir, line 15: Should the `operator` also be a signature ?

brian-weir (approved):

daml/Marketplace/Clearing/Market/Model.daml
```diff
@@ -0,0 +1,51 @@
...
+    controller provider can
+      FairValueCalculationRequest_Ack : ()
+        do return ()
+
+template ManualFairValueCalculation
+  with
+    provider : Party
```
brian-weir, line 22: Should these templates have the `operator : Party` as part of them ?

ui2/src/pages/listing/Listing.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const listing = listings.find(a => a.contractId === cid);
+  const service = services.find(s => s.payload.customer === party);
+
+  const { contracts: fairValueContracts, loading: fairValuesLoading } = useStreamQueries(FairValue);
+  const { contracts: clearedMarketServices } = useStreamQueries(ClearedMarketService);
+  const clearedMarketService = clearedMarketServices.find(
+    cms => cms.payload.customer === listing?.payload.customer
+  );
+  console.log(clearedMarketService);
```
Sarahb-da, line 83: extra log

ui2/src/pages/listing/Listing.tsx
```diff
@@ -0,0 +1,185 @@
...
+            )}
+            <Button floated="left" className="ghost" onClick={() => requestDisableDelisting()}>
+              Disable
+            </Button>
+          </Button.Group>
+        </div>
+      </Tile>
+      {!!listing && (
+        <Tile header={<h4>Details</h4>}>
+          <StripedTable
```
Sarahb-da, line 126: u could add `loading` to line 73 and use it here to give this table a loading state

ui2/src/pages/listing/Listings.tsx
```diff
@@ -18,49 +25,88 @@ export const ListingsTable: React.FC<Props> = ({ services, listings }) => {
...
+          'Listing ID',
+          'Calendar ID',
+          'Traded Asset',
+          'Traded Asset Precision',
+          'Quoted Asset',
+          'Quoted Asset Precision',
+          'Fair Value',
+          'Action',
+        ]}
+        rowsClickable
```
Sarahb-da, line 70: Sadly I dont think this will work - the row being clickable overlaps with the button in the row. Making the click functionality for the button possibly not work.. its pretty poor design on my part... in order to have clickable rows and buttons there needs to be some kind of click listener added. So the DOM knows if the user clicked the row, or the button in the row.

Sarahb-da: basically `rowsClickable` only works for tables with no buttons in them

ui2/src/pages/listing/ManualCalculationRequests.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <Button.Group floated="right">
+              <ModalFormErrorHandled onSubmit={() => applyCalculation(fv)} title="Calculate">
+                <Form.Input
+                  label="Amount"
+                  placeholder="Select..."
+                  required
+                  number
+                  min={1}
+                  value={valueAmount}
+                  onChange={(_, change) => setValueAmount(change.value as string)}
```
Sarahb-da, line 50: does the `rowsClickable` prop mess up the `input`? curious

Sarahb-da: u actually dont need it anyway bc the row has no `onClick` function

Sarahb-da (approved):
just a few minor comments but looks great

ui2/src/pages/listing/ManualCalculationRequests.tsx
alexgraham-da: the `Input` is in the `ModalFormErrorHandled` so it only pops up when the modal does. row doesn't need to be clickable though, so I'll remove it

alexmatson-da (approved):

daml/Marketplace/Clearing/Market/Model.daml
alexgraham-da: I can add this. I'm curious though, I've noticed on multiple models I've worked on that the `operator` is a `signatory` on the majority of contracts, even ones the `operator` does not need to perform any action on. What is the general purpose of this? Is it to provide visibility to the ledger administrator (and wouldn't this be a possible privacy concern?). Or maybe to disallow a party from archiving contracts without using a `choice`?

ui2/src/pages/listing/Listings.tsx
alexgraham-da: ah OK - I can move that button to the listing's page and look into adding a listener like that (in another PR).

daml/Marketplace/Clearing/Market/Model.daml
```diff
@@ -0,0 +1,51 @@
...
+
+import DA.Finance.Types
+import DA.Next.Set (Set)
+
+template FairValueCalculationRequest
+  with
+    operator : Party
+    provider : Party
+    customer : Party
+    optListingIds : Optional [Text]
```
lucianojoublanc-da, line 11: Probably worth `newtype`ing this. How are we currently identifying listings?

daml/Marketplace/Clearing/Market/Model.daml
```diff
@@ -0,0 +1,51 @@
...
+    provider : Party
+    customer : Party
+    listingId : Text
+    currency : Id
+    upTo : Time
+    observers : Set Party
+  where
+    signatory provider
+
+    controller customer can
```
lucianojoublanc-da, line 31: Doesn't make sense to me from a business point of view that the customer is valuing their own positions... Maybe related to @brian-weir 's comment above.

daml/Marketplace/Clearing/Market/Service.daml
```diff
@@ -16,8 +20,18 @@ template Service
     key (operator, provider, customer) : (Party, Party, Party)
     maintainer key._1
 
-    controller customer can
+    nonconsuming choice RequestFairValues : ContractId FairValueCalculationRequest
```
lucianojoublanc-da, line 23:
```suggestion
    nonconsuming choice RequestFairValue : ContractId FairValueCalculationRequest
```
Is this a typo?

triggers/daml/ClearingTrigger.daml
```diff
@@ -146,3 +162,52 @@ collectDeposits restFunds (used, (nextDeposit::remaining))
...
+--    2. If there are no trades, the most recent 'FairValue'
+--    3. If there are no 'FairValue', request manual calculation
+--
+-- TODO: Allow user to specify a 'from' Time interval before which trades and
+-- fairvalues will be ignored.
+handleFairValueCalculation : Party -> Id -> Time -> ContractPair Listing.T -> TriggerA ClearingState ()
+handleFairValueCalculation party currency upTo (marketPairCid,l@(Listing.Listing{..})) = do
+  debug ("calculating FVR: " <> l.listingId)
+
+  fairValues <- sortOn ((.upTo) . snd)
```
lucianojoublanc-da, line 193:
```suggestion
  fairValues <- sortOn (upTo . snd)
```

triggers/daml/ClearingTrigger.daml
```diff
@@ -146,3 +162,52 @@ collectDeposits restFunds (used, (nextDeposit::remaining))
...
+        (Some ids) -> filterField (.listingId) (`Set.member` ids) allListings
+        None       -> allListings
+
+  forA_ listings (handleFairValueCalculation party currency upTo)
+  emitExerciseCmd requestCid Clearing.FairValueCalculationRequest_Ack
+  return ()
+
+-- |Calculates a FairValue for a single derivative.
+-- Order of precedence:
+--    1. The price of the most recent trade before the specified 'upTo'
```
lucianojoublanc-da, line 183: This doesn't sound right. Shouldn't you be taking price _after_ the specified date?

triggers/daml/ClearingTrigger.daml
```diff
@@ -146,3 +162,52 @@ collectDeposits restFunds (used, (nextDeposit::remaining))
...
+  case getBestPrice relevantTrades fairValues of
+    (Some bp) -> dedupCreate $ Clearing.FairValue with provider = party, price = bp; ..
+    None      -> dedupCreate $ Clearing.ManualFairValueCalculation with provider = party,  ..
+
+  return ()
+
+  where
+    getBestPrice : [ContractPair Clearing.ClearedTradeSide] -> [ContractPair Clearing.FairValue] -> Optional Decimal
+    getBestPrice ts fvs
+      | not (null ts)  = Some . (.price) . (.execution) . snd . head $ ts
```
lucianojoublanc-da, line 211:
```suggestion
      | not (null ts)  = Some . price . execution . snd . head $ ts
```
Does that not work?

lucianojoublanc-da (commented):
I'd like to understand better in what context this is used. Because this is quite a sensitive thing - getting this wrong can lead to large realized losses. It's particularly problematic when you have thinly traded securities, where the price isn't representative at all, so I'm a bit worried about this logic of looking at the last price, or similar ... :grimacing:

daml/Marketplace/Clearing/Market/Model.daml
alexgraham-da: The customer here is the listing agent, not a clearinghouse member, this is the same `Service` that the listing agent uses to request the clearinghouse cover their market.  Does it make more sense to you that way? It used to be the exchange handling any manual valuations, and since that is now split out into multiple services, the listing part of it made the most sense to me.

In the end I think this will fit best somewhere in the listing service directly, so any customer can get fair values of instruments (regardless of if they are cleared or not).

daml/Marketplace/Clearing/Market/Model.daml
alexgraham-da: Elsewhere they've been `listingId: Text`, but I think using `Id` might be better so you can `key` the `Listing` with the `Id`. Currently if you're trying to do a `fetchByKey` of a listing, you wouldn't know who the provider/customer is to do the lookup.

triggers/daml/ClearingTrigger.daml
alexgraham-da: It does not...`(.price)` is a field accessor, I believe it's the only way of doing `contract.field` as a single function, at least for a `template`.

daml/Marketplace/Clearing/Market/Service.daml
alexgraham-da: It was not since you are able to use this choice to request `FairValue`s for multiple listings. Now that I'm looking at it, it might make some sense to have multiple choices - `RequestFairValue` for one listing, `RequestFairValues` for multiple listings, and `RequestAllFairValues` for all listings instead of having the `Optional` that doesn't really suggest the actual functionality (`None` meaning all listings)...

triggers/daml/ClearingTrigger.daml
alexgraham-da: `upTo` was how the requirement was given, the idea being "I would like the latest fair value, ignoring anything after 5pm". It might make sense to also have an optional `after : Time` to allow for a range.

triggers/daml/ClearingTrigger.daml
alexgraham-da: See https://github.com/digital-asset/da-marketplace/pull/300#discussion_r625891796

triggers/daml/ClearingTrigger.daml
alexgraham-da: Actually, I think you _can_ use the accessors as functions without the `.`, but since we aren't bothering making field names unique, it won't infer which function to use...so since multiple templates in the `Clearing` namespace have a `price` field, you get an ambiguous occurrence error.

alexgraham-da:
> I'd like to understand better in what context this is used. Because this is quite a sensitive thing - getting this wrong can lead to large realized losses. It's particularly problematic when you have thinly traded securities, where the price isn't representative at all, so I'm a bit worried about this logic of looking at the last price, or similar ... 😬

@lucianojoublanc-da - this is all used as a way for the clearinghouse (currently only Baymarkets) to have a way of calculating mark to markets and margins. The algorithm for determining the fair values was a simplistic one to use as a POC and can be changed however is necessary. I think a good way to handle this would be to allow the customer to choose from different methods of calculating fair values, maybe even in a composable way similar to the contingent claims library.

daml/Marketplace/Clearing/Market/Model.daml
alexgraham-da: I am addressing this is a separate PR

daml/Marketplace/Clearing/Market/Service.daml
alexgraham-da: I split the choice as mentioned above.

daml/Marketplace/Clearing/Market/Model.daml
lucianojoublanc-da: I would avoid using 'customer' as it can mean many things. Typically it would not be a 3rd party with no affiliations. Does either 'issuer' or 'exchange' fit the description?

triggers/daml/ClearingTrigger.daml
```diff
@@ -146,3 +162,52 @@ collectDeposits restFunds (used, (nextDeposit::remaining))
...
+--    3. If there are no 'FairValue', request manual calculation
+--
+-- TODO: Allow user to specify a 'from' Time interval before which trades and
+-- fairvalues will be ignored.
+handleFairValueCalculation : Party -> Id -> Time -> ContractPair Listing.T -> TriggerA ClearingState ()
+handleFairValueCalculation party currency upTo (marketPairCid,l@(Listing.Listing{..})) = do
+  debug ("calculating FVR: " <> l.listingId)
+
+  fairValues <- sortOn ((.upTo) . snd)
+            <$> filterQuery @Clearing.FairValue (.listingId) (== l.listingId)
```
lucianojoublanc-da, line 197: Btw not sure this `filterQuery` is helping at all; think that in-lining this with a lambda would be much more legible.

triggers/daml/ClearingTrigger.daml
lucianojoublanc-da: So what if
- the trade was at 9 am this morning? Will that be ignored because it's after 5 pm yesterday?
- the last trade was three days ago? (This happens in real life).

> upTo was how the requirement was given,

I'd like to understand if this was an internal requirement / Exberry / a client. Maybe we can chat off-line.

triggers/daml/ClearingTrigger.daml
alexgraham-da: I personally like the way it looks since if you read through the lambdas it reads as an expression (`Template.fieldName == variable)` but if you're strongly against it I can change. I find the lambdas pattern matching the tuples all over these triggers ugly, but it's readable in that it is exactly how it looks, so that's fine.

triggers/daml/ClearingTrigger.daml
alexgraham-da: This was an Eric requirement, but we can chat about it. For the first one, yes, if you specified 5pm yesterday then the trade was at 9am this morning will be ignored by design, since you're interested in doing calculations based on what happened yesterday. For the second, also yes based on this particular algorithm (which is overly simple and can be changed) since that is the most recent traded value.

daml/Marketplace/Clearing/Market/Model.daml
alexgraham-da: I was going with `customer` to match the service but I would prefer `exchange` anyway.

brian-weir: Almost all our services have `operator, provider, customer` parties as part of their template and also the template's signature. It makes it clear which party is the producer and which is the consumer of said services. Plus, once we get generics (or what they are calling 'interfaces') back into Daml, it should (in theory...) be easy to extract these details out of all our services (along with the offer/request templates) if we're consistent in the code base.

I do understand your point that `exchange` provides more context to the expected party... would something like the below which addresses both concerns help ? 

> let exchange = customer
>
> controller exchange can

daml/Marketplace/Clearing/Market/Model.daml
brian-weir: Good question - this main use of this pattern is to enforce that contracts are created only via service `choice`s which require the signature from the `operator` of that Marketplace. In other words, a `provider` (or a `customer` or even an `operator`...) can't simply create contracts outside of the "approved" service context, they must follow the explicit workflow provided by a service.

Regarding the privacy concern of this pattern, it's correct that an `operator` will have visibility of all the contracts they are a signatory to. Whether this is an issue or not likely depends on how you view the the `operator` which runs such a Marketplace. Should this be viewed as `admin` access (akin to a database super user access) which can view and monitor an entire Marketplace (makes it easier to audit, regulatory reporting, etc.) ? Or, should templates only have parties which are strictly required be allowed ? I think there's a trade off to be made here and currently we're enforcing having the correct signatures over privacy.

@GeorgSchneider , do you have view on this ?

GeorgSchneider: yep, having the operator on there is merely to be able to restrict "valid" contracts in an operator-centric marketplace. That is, a given marketplace (operator) will only accept contracts that are signed by the operator, such that different parties cannot act "outside" of the provided service framework. In the decentralized operating model, the services would be "self-signed" with the providers as the respective operators. Re privacy, it is by design that an operator-centric marketplace would give full visibility for the operator. The decentralized mode is one that would retain certain privacy properties.

In summary I'd suggest to add the operator here as the signatory, assuming that the service this originates from does have the operator signature on it so it can be done easily.

alexgraham-da: This makes a lot of sense, thanks all for the explanations. I actually noticed when adding the `operator` that I was required then to create these contracts in the trigger via the `Service`, so that gave me a nice practical example of why this pattern is useful 😃 .

PR #301, opened 2021-05-03
Use settlement instructions instead of trasferring

gaborh-da:


daml/Marketplace/Custody/Model.daml
```diff
@@ -1,4 +1,14 @@
...
+module Marketplace.Custody.Model (
+    OpenAccountRequest(..)
+  , OpenAllocationAccountRequest(..)
+  , CloseAccountRequest(..)
+  , CloseAllocationAccountRequest(..)
+  , CreditAccountRequest(..)
+  , DebitAccountRequest(..)
+  , TransferDepositRequest(..)
+  , LifecycleRequest(..)
+  , Process(..)
```
lucianojoublanc-da, line 10:
```suggestion
```

Also please remove comment line110.

daml/Marketplace/Custody/Model.daml
```diff
@@ -97,11 +108,12 @@ template LifecycleRequest
     signatory operator, provider, customer
 
     -- TODO: this shouldn't be exported; it's only called from CustodyService.Lifecycle
-    choice Process : [ ContractId AssetDeposit ]
+    choice Process : (ContractId AssetDeposit, [ContractId SettlementInstruction])
       with
           investor: Party
           safekeepingDepositCid: ContractId AssetDeposit
           fixings: Map Text (Map Date Decimal) --TODO these should be looked up from a contract on-ledger, not provided by the custodian!
+          uniquePayoutId: Text
```
lucianojoublanc-da, line 116: So you decided that ultimately this is provided off-ledger, right? cc @georg-da @brian-weir you agree this is ok?

lucianojoublanc-da (commented):
Looks good to me. I'd like one of the other guys to also peek at it, especially the thing with trade ids.

daml/Marketplace/Custody/Model.daml
gaborh-da: Service.daml, line 140:
```
Custody.Process
```
I cannot remove that.

daml/Marketplace/Custody/Model.daml
gaborh-da: Yes.

daml/Marketplace/Custody/Model.daml
lucianojoublanc-da: Ah, ok. I remember now we refactored from a single file to multiple `Model/Service` files, so this no longer applies. Sorry, my bad.

daml/Marketplace/Custody/Model.daml
```diff
@@ -145,3 +174,10 @@ createNewVersionOfDescription description postAssetVersion serializedRemainingCl
     claims = serializedRemainingClaims
  where
   adjustedId = description.assetId with version = postAssetVersion
+
+createTradeId : Id -> Int -> Text -> Id
+createTradeId assetId transferNumber uniqueId =
+  Id with
+    signatories = assetId.signatories
+    label = assetId.label <> " " <> show transferNumber <> "th transfer, ID: " <> uniqueId
```
brian-weir, line 182: what is `th transfer` mean ?

daml/Marketplace/Custody/Model.daml
brian-weir: So the investor is expected to provide this `uniquePayoutId` as they call are calling this choice ?

daml/Marketplace/Custody/Model.daml
gaborh-da: "4th transfer", there can be many transfers / payouts when process/lifecycle (the choice) is called.

daml/Marketplace/Custody/Model.daml
gaborh-da: This choice is called by the "provider". That party is expected to provide a unique payout ID. That ID can be implemented in many ways, see Tests/Bond.daml - the date is used, also Tests/BinaryOption.daml - some "internal" (in this case imaginary) ID is used. One can also implement a counter contract and use that to provide the ID.

daml/Marketplace/Custody/Model.daml
brian-weir: ok, that makes sense

daml/Marketplace/Custody/Service.daml
```diff
@@ -130,11 +131,12 @@ template Service
           Custody.TransferDepositRequest{accountId; transfer} <- fetchAndArchive transferDepositRequestCid
           exerciseByKey @AssetSettlementRule accountId transfer
 
-      nonconsuming Lifecycle : [ContractId AssetDeposit]
+      nonconsuming Lifecycle : (ContractId AssetDeposit, [ContractId SettlementInstruction])
         with
           lifecycleRequestCid : ContractId Custody.LifecycleRequest
           safekeepingDepositCid: ContractId AssetDeposit
           fixings: Map Text (Map Date Decimal) --TODO these should be looked up from a contract on-ledger, not provided by the custodian!
+          uniquePayoutId: Text
```
alexgraham-da, line 139: following the style from the rest of the marketplace model, type declarations should be `varName : Type`, not `varName: Type`.

alexgraham-da: actually, there does seem to be some inconsistency across the model, however this file all has the space before the `:` so I would go with that here

daml/Marketplace/Custody/Model.daml
brian-weir: I'm just realising that `controller investor` was already there but you've added `masterAgreementSignatories = [provider, customer, investor]`. Thus, `provider` and `investor` is actually the same party ?

gaborh-da: Ah I overlooked `controller investor` but was referring to the following in my previous comment:
https://github.com/digital-asset/da-marketplace/pull/301/files#diff-fbb5e9937b9cb7cb0846524c164f5290d66dc583bce10491a9ea6ac8b094e5a3R129
(see that the provider aka. bank does the lifecycling).

It seems that they can be the same party in some cases.
When defining masterAgreementSignatories, I tried to include all stakeholders (I didn't add or modify either provider, customer or investor).

daml/Marketplace/Custody/Service.daml
gaborh-da: Should I modify 136-138 as well?

alexgraham-da: yeah that would be great, thank you 👍

alexgraham-da (approved):
Looks like @brian-weir is still looking for some clarification on the `Process` choice, but otherwise it looks good to me 👍

daml/Marketplace/Custody/Model.daml
brian-weir: Cool 👍 

We'll tidy this up in another PR with @lucianojoublanc-da. It looks like the customer of the Custody service requests `RequestLifecycle` which creates a `Custody.LifecycleRequest` contract with the `provider` and `customer` details from the service and thus `investor: Party` on the `LifecycleRequest.Process` choice can only be party which is a signatory to that contract.

brian-weir (approved):

PR #303, opened 2021-05-04
Add Allocation accounts to the UI

brian-weir:
Added Allocation accounts to the UI as part of the Wallet screen. Also removed the public party as an observer to newly created accounts in the UI.

### DIFF
--- a/daml.yaml
+++ b/daml.yaml
@@ -6,9 +6,6 @@ parties:
   - Alice
   - Bob
   - Operator
-  - Custodian
-  - BtcIssuer
-  - UsdtIssuer
   - Public
   - Exchange
   - Ccp
--- a/daml.yaml
+++ b/daml.yaml
@@ -14,6 +14,7 @@ parties:
   - Ccp
   - Broker
   - Bank
+  - Issuer
 version: 0.1.8
 dependencies:
   - daml-prim
--- a/ui2/src/pages/custody/Account.tsx
+++ b/ui2/src/pages/custody/Account.tsx
@@ -1,4 +1,4 @@
-import React, {useMemo, useState} from 'react';
+import React, { useMemo, useState } from 'react';
 import { useLedger, useParty } from '@daml/react';
 import { useStreamQueries } from '../../Main';
 import { AssetSettlementRule } from '@daml.js/da-marketplace/lib/DA/Finance/Asset/Settlement';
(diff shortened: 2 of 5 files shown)


ui2/src/pages/custody/Account.tsx
```diff
@@ -28,9 +29,19 @@ const AccountComponent: React.FC<RouteComponentProps & ServicePageProps<Service>
   const cid = contractId.replace('_', '#');
 
   const { contracts: accounts, loading: accountsLoading } = useStreamQueries(AssetSettlementRule);
+  const { contracts: allocatedAccounts, loading: allocatedAccountsLoading } = useStreamQueries(AllocationAccountRule);
   const { contracts: deposits, loading: depositsLoading } = useStreamQueries(AssetDeposit);
   const { contracts: assets, loading: assetsLoading } = useStreamQueries(AssetDescription);
 
+  const allAccounts = useMemo(
+    () => accounts.map(a => {
+      return { account: a.payload.account, contractId: a.contractId.replace('#', '_') }
```
alexmatson-da, line 38: I've noticed this before as well -- why replace `#` with `_`?

daml.yaml
```diff
@@ -14,6 +14,7 @@ parties:
   - Ccp
   - Broker
   - Bank
+  - Issuer
```
alexgraham-da, line 17: I think we can get rid of `BtcIssuer`, `UsdtIssuer` and `Custodian` from here as well, those were leftover from the old model's setup script

alexmatson-da (approved):

ui2/src/pages/custody/Account.tsx
```diff
@@ -15,6 +15,7 @@ import { usePartyName } from '../../config';
 import StripedTable from '../../components/Table/StripedTable';
 import { ServicePageProps } from '../common';
 import { ArrowLeftIcon } from '../../icons/icons';
+import {AllocationAccountRule} from "@daml.js/da-marketplace/lib/Marketplace/Rule/AllocationAccount/module";
```
alexgraham-da, line 18: did you run `yarn format`? should have caught the missing spaces between the `{}`

alexmatson-da: oh and a nit pick - could you remove the `/module` at the end of the import? I've noticed VS Code sometimes puts that in automatically, and it's not required

alexgraham-da (approved):

daml.yaml
brian-weir: Good points - I'll remove them with the next commit.

ui2/src/pages/custody/Account.tsx
brian-weir: That's a good question - I presume its because we're using a `<HashRouter>` in our code base that if the `Contract Id` contains a `#` then it will break the router as there'll be two hashes in the URL. To be transparent, I haven't tested this theory. This logic was originally implemented by @GeorgSchneider and I've simply ensured its consistent when generating URLs with `Contract Id`s.

ui2/src/pages/custody/Account.tsx
brian-weir: I'll fix both points on the next commit - well spotted!! 👀

ui2/src/pages/custody/Account.tsx
GeorgSchneider: Indeed, there's a problem with the router when there's a `#` in the URL. Hence, the replacement here and back again in the page being navigated to.

PR #306, opened 2021-05-05
Point to new OSS claims repo

lucianojoublanc-da:
This points to the new open-source repo under the digital-asset
organisation. The hashes differ as the history was re-written,
but the commit is the same.


brian-weir (approved):
lgtm

PR #316, opened 2021-05-06
Missing service errors

alexgraham-da:
added a guard for pages that require a service to show a popup giving them the option to request a service if they do not currently have it


ui2/src/pages/error/ServiceRequired.tsx
```diff
@@ -0,0 +1,259 @@
+import React, { useEffect, useState, useMemo } from 'react';
```
alexgraham-da, line 1: This page shares a lot of code with `Landing.tsx`, the actual modal they both use can probably be refactored into one component

alexmatson-da (approved):
Looks good to me

PR #317, opened 2021-05-06
add links in manage pages

alexgraham-da:
adds action links in the manage page

### DIFF
--- a/ui2/src/pages/network/Actions.tsx
+++ b/ui2/src/pages/network/Actions.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import Tile from '../../components/Tile/Tile';
 import { Button } from 'semantic-ui-react';
-import { NavLink } from 'react-router-dom';
+import { Link } from 'react-router-dom';
 
 type Action = {
   path: string;
@@ -18,11 +18,11 @@ export const ActionTile: React.FC<Params> = ({ actions }) => {
       <div className="action-row">
         <Button.Group>
           {actions.map(a => (
-            <NavLink to={a.path}>
+            <Link to={a.path}>
               <Button floated="left" className="ghost">
                 {a.label}
               </Button>
-            </NavLink>
+            </Link>
           ))}
         </Button.Group>
       </div>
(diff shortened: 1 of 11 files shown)


ui2/src/pages/setup/Offer.tsx
```diff
@@ -52,27 +52,32 @@ const Offer: React.FC<{ service: ServiceKind }> = ({ service }) => {
...
       }
       case ServiceKind.LISTING: {
         setOffer(ListingOffer);
         setChoice(TradingRole.OfferListingService);
+        break;
       }
       case ServiceKind.CLEARING: {
         setOffer(CustodyOffer);
         setChoice(ClearingRole.OfferClearingService);
+        break;
```
alexmatson-da, line 80: Nice, also add a `default` case that throws an Error (mainly just because switch statements without a `default` cause a build warning)

ui2/src/pages/network/Actions.tsx
```diff
@@ -0,0 +1,31 @@
...
+  actions: Action[];
+};
+
+export const ActionTile: React.FC<Params> = ({ actions }) => {
+  return (
+    <Tile header={<h4>Actions</h4>}>
+      <div className="action-row">
+        <Button.Group>
+          {actions.map(a => (
+            <NavLink to={a.path}>
```
alexmatson-da, line 21: This should probably be a `<Link>` rather than a `<NavLink>`. I've always been prone to favor NavLinks in the past, but they're not necessary unless you need the CSS classname for active styling :grin:

alexmatson-da (approved):
This looks very helpful, thanks!

ui2/src/pages/network/Distribution.tsx
```diff
@@ -40,21 +41,24 @@ export const DistributionServiceTable = () => {
...
-            getName(c.payload.operator),
-            getName(c.payload.provider),
-            getName(c.payload.customer),
-            party === c.payload.provider ? 'Provider' : 'Consumer',
-          ],
-        };
-      })}
-    />
+    <>
+      <ActionTile actions={[{ path: '/app/distribution/new', label: 'New Distribution' }]} />
```
alexmatson-da, line 45: Should this path be `/app/setup/distribution/new/auction`?

I wonder if its time to have a global list of app paths somewhere, too...

alexgraham-da: Ah yeah, I see now that path in `App.tsx` uses the same component as `Auction` just named differently...will change. Agreed that a global app path list would be good!

PR #320, opened 2021-05-11
Allow fields on notifications

alexmatson-da:


brian-weir (approved):

alexgraham-da (commented):
I'm late to the party, but very nice!

PR #331, opened 2021-05-18
move ui2 to ui

alexmatson-da:
this diff is simply the result of deleting `ui/` and renaming `ui2/` to `ui/`

I also updated the Makefile to point to the new directory, so building should still work.


brian-weir (approved):

alexgraham-da (approved):
RIP

PR #341, opened 2021-05-19
Add source url and version bump..

Sarahb-da:


alexgraham-da (approved):
looks good once it's changed to `0.2.0`

also it looks like `ui2/src/pages/issuance/Issuances.tsx` snuck back in by accident a few commits ago if you want to nuke it in this PR

PR #342, opened 2021-05-20
various ui improvements

Sarahb-da:
Tries to unify some language, and use icons where possible to indicate actions more clearly 

Creates common page layouts 

(the branch is incorrectly named - this PR does not do that.)


ui/src/App.tsx
```diff
@@ -427,7 +427,12 @@ const AppComponent = () => {
   };
 
   const path = useLocation().pathname;
-  const currentEntry = entriesToDisplay.find(entry => path.startsWith(entry.path));
+
+  const currentEntry = entriesToDisplay.find(entry => path.startsWith(getBaseSegment(entry.path)));
+
+  function getBaseSegment(segment: string) {
+    return [segment.split('/')[0], segment.split('/')[1], segment.split('/')[2]].join('/');
+  }
```
Sarahb-da, line 435: this makes it so `additional routes` fall under the same menuTitle as their parent segments. ex. Anything deeper than `app/custody` will have the `custody` menu title and so on

ui/src/components/InputDialog/InputDialog.scss
```diff
@@ -0,0 +1,57 @@
+@use '../../themes/variables' as *;
```
Sarahb-da, line 1: this will b the common layout of input forms.

ui/src/pages/clearing/Member.tsx
```diff
@@ -19,9 +19,8 @@ import StripedTable from '../../components/Table/StripedTable';
 import MarginCallModal from './MarginCallModal';
```
Sarahb-da, line 19: I just move stuff around in this file, not changing any content

ui/src/pages/listing/New.tsx
```diff
@@ -107,115 +108,114 @@ const NewComponent: React.FC<RouteComponentProps & ServicePageProps<Service>> =
   const { step, placeholder } = preciseInputSteps(+tradedAssetPrecision);
 
   return (
-    <div className="listing">
```
Sarahb-da, line 110: just moving stuff around here, not changing anything

ui/src/pages/origination/NewBaseInstrument.tsx
```diff
@@ -10,12 +10,12 @@ import { Date as DamlDate } from '@daml/types';
...
-import Tile from '../../components/Tile/Tile';
 import FormErrorHandled from '../../components/Form/FormErrorHandled';
 import { Button, Form, Header } from 'semantic-ui-react';
 import classNames from 'classnames';
-import { IconCircledCheck, LockIcon, PublicIcon } from '../../icons/icons';
+import { IconCircledCheck, LockIcon, PublicIcon, IconClose } from '../../icons/icons';
 import { publicParty } from '../../config';
+import BackButton from '../../components/Common/BackButton';
 
 const NewBaseInstrumentComponent = ({ history }: RouteComponentProps) => {
```
Sarahb-da, line 20: just moving stuff around here, not changing anything

ui/src/pages/origination/NewBinaryOption.tsx
```diff
@@ -120,108 +123,105 @@ const NewBinaryOptionComponent = ({ history }: RouteComponentProps) => {
     getContentAnchorEl: null,
   };
   return (
-    <div className="new-binary-option">
```
Sarahb-da, line 123: just moving stuff around here, not changing anything

ui/src/pages/origination/NewConvertibleNote.tsx
```diff
@@ -160,136 +162,119 @@ const NewConvertibleNoteComponent = ({ history }: RouteComponentProps) => {
     getContentAnchorEl: null,
   };
   return (
-    <div className="new-convertible-note">
```
Sarahb-da, line 163: just moving stuff around here, not changing anything

ui/src/components/Table/StripedTable.scss
```diff
@@ -53,42 +84,42 @@
...
+    //         }
+
+    //         .ui.button {
+    //           display: unset;
+    //           margin: unset;
+    //         }
+    //       }
+    //     }
+    //   }
+    // }
```
brian-weir, line 123: Can you remove these comments ?

ui/src/components/Table/StripedTable.tsx
```diff
@@ -48,6 +43,13 @@ const StripedTable = (props: {
     }
   }, [activePage, rows, rowsPerPage]);
 
+  if (activePageRows.length === 0) {
+    return (
+      <div className="striped-table empty">
+        <p className="p2">This table contains on data...yet..</p>
```
brian-weir, line 45: I think you mean "This table doesn't contain data yet..." or something like that ?

Sarahb-da: pshh yes. Haha. Thanks!

ui/src/pages/custody/Accounts.tsx
```diff
@@ -42,10 +45,42 @@ const AccountsComponent: React.FC<RouteComponentProps & Props> = ({
...
+        if (newObservers.length > 0) {
+          addSignatoryAsDepositObserver(d, newObservers);
+        }
+      })
+    );
+  };
+
+  useEffect(() => {
+    updateDeposits(3);
+  }, []);
```
brian-weir: I think we'll have to come up with another solution for this problem as this solution will cause issues for any services that have a stored `AssetDeposit` contract id on their contract (Ie, Trading, Custody, Distribution, etc.).  


Looking at Eric's messages on Slack
> it's much better if the issuer has their own cap table and holds all shares themselves. and can even have customers hold assets on the issuer's cap table.
 
I'm thinking we can solve this by having the issuer be a custody provider and clients of the issuer can open accounts at said issuer. This would mean the cap table and the shares would both be owned by the issuer and solve Erics problem. What do you think ?

ui/src/pages/distribution/auction/Auction.tsx
```diff
@@ -47,6 +50,34 @@ export const Auction: React.FC<RouteComponentProps & Props> = ({
...
+      deposits.map(d => {
+        const newObservers = damlSetValues(d.payload.asset.id.signatories).filter(
+          signatory => !d.observers.includes(signatory)
+        );
+        if (newObservers.length > 0) {
+          addSignatoryAsDepositObserver(d, newObservers);
+        }
+      })
+    );
+  };
```
brian-weir: Is this also so the issuer is added as an observer to the `AssetDeposit` ?

ui/src/pages/custody/Accounts.tsx
Sarahb-da: yes I removed this from the pr, should be updated as of my latest commit

Sarahb-da: Because it deserves its own pr, definitely

Sarahb-da: I agree I think this needs to be redone, and I like that approach !

ui/src/pages/distribution/auction/Auction.tsx
Sarahb-da: See my reply above ^

brian-weir (approved):
lgtm!! Thanks @Sarahb-da !!

PR #345, opened 2021-05-21
update scripts/makefile to handle pre-release semantic version schemes

alexmatson-da:


.circleci/config.yml
```diff
@@ -137,8 +137,16 @@ workflows:
     jobs:
       - daml_test:
           daml_sdk_version: "1.10.0"
```
alexgraham-da, line 139: Should these be `1.12.0`?

alexmatson-da: Yep.. I'll update that

alexgraham-da (approved):
nice, lgtm!

PR #346, opened 2021-05-21
Fix warnings

alexmatson-da:
This goes through and resolves all of the simple build warnings. 

Stuff like unused variables, `==` vs `===`, etc. 

There are still a handful of warnings remaining, but they are somewhat more complex to solve. Many of them are useEffect dependency warnings, and changing those arrays can have unexpected side effects, so I'm going to go through those more carefully.


Sarahb-da (approved):
ty!

PR #349, opened 2021-05-24
Add observers to issuance and credit requests, and pull through for all Asset Deposits

Sarahb-da:
We can make this optional for the user, but for now I just added the observers in all requests (the field is not part of the form that a user fills out to create an issuance or asset deposit)


daml/Marketplace/Custody/Service.daml
```diff
@@ -114,8 +116,9 @@ template Service
         with
           creditAccountRequestCid : ContractId Custody.CreditAccountRequest
         do
-          Custody.CreditAccountRequest{accountId; asset} <- fetchAndArchive creditAccountRequestCid
-          exerciseByKey @AssetSettlementRule accountId AssetSettlement_Credit with ctrl = provider; ..
+          Custody.CreditAccountRequest{accountId; asset; observers = assetObservers} <- fetchAndArchive creditAccountRequestCid
```
brian-weir, line 119: Whats the reason behind `observers = assetObservers` here ?

PR #351, opened 2021-05-25
clean up unused code / unused exports

alexmatson-da:


ui/src/automation.ts
```diff
@@ -2,6 +2,9 @@ import { dablHostname, deploymentMode, DeploymentMode, httpBaseUrl, ledgerId } f
 import { computeCredentials } from './Credentials';
 
 export const TRIGGER_HASH = process.env.REACT_APP_TRIGGER_HASH;
+
+// Do we want to keep this?
+// ts-prune-ignore-next
 export const EXBERRY_HASH = process.env.REACT_APP_EXBERRY_HASH;
```
alexmatson-da, line 8: I've noticed here (and in other places throughout) there are some vars that are unused but look like we might use in the future. 

So I just left comments on them instead of deleting, if you see one that you think we can delete let me know!

ui/src/components/BottomDrawer/BottomDrawer.tsx
```diff
@@ -1,61 +0,0 @@
-import React from 'react';
```
Sarahb-da, line 1: ive always wondered about this one lol

ui/src/context/MessagesContext.tsx
```diff
@@ -80,6 +80,8 @@ function useDisplayErrorMessage() {
   return context.displayErrorMessage;
 }
 
+// Do we want to keep this?
+// eslint-disable-next-line
 function useDisplaySuccessMessage() {
```
Sarahb-da, line 85: what do u think? I think we can actually get rid of it. Success messages are kinda a redundant think anyway

ui/src/context/RequestsContext.tsx
```diff
@@ -37,6 +37,8 @@ export enum ServiceKind {
   REGULATOR = 'Regulator',
 }
 
+// Do we want to keep this?
+// ts-prune-ignore-next
```
Sarahb-da, line 41: I would say leave it for now

ui/src/context/RolesContext.tsx
```diff
@@ -118,9 +120,9 @@ function useRoleKinds(): Set<RoleKind> {
 function useRolesContext() {
   const context = React.useContext<RolesState>(RolesStateContext);
   if (context === undefined) {
-    throw new Error('useRoles must be used within a RolessProvider');
+    throw new Error('useRoles must be used within a RolesProvider');
   }
   return context;
 }
 
-export { RolesProvider, useRolesContext, useRoleKinds };
```
Sarahb-da, line 126: I feel like if we do keep it we should leave it exported? So that this file doesnt have to change if it is used in the future?

ui/src/pages/origination/Instruments.tsx
```diff
@@ -50,19 +48,3 @@ export const InstrumentsTable: React.FC = () => {
     </div>
   );
 };
-
-const InstrumentsComponent: React.FC<RouteComponentProps> = ({ history }: RouteComponentProps) => {
```
Sarahb-da, line 54: Yay! I am glad these are going away..

Sarahb-da (approved):
A few comments as to if we keep certain things - Otherwise great.

PR #353, opened 2021-05-25
Filtered request offer fields

alexgraham-da:
- Add asserts to trading and clearing account creation to make sure the accounts are visible or nominated to the provider
- Filter accept offer and request offer fields in the UI to show only viable accounts
- Reduce character amount of error message displays (useful for giant Daml grpcio errors)
- Add disabled to `InputDialog`
- Fix service offer dialog for trading
- Clearing offer modal that can automatically request accounts
- Add observers field to new accounts


ui/src/pages/clearing/ClearingOfferModal.tsx
```diff
@@ -0,0 +1,175 @@
...
+  OpenAccountRequest,
+  OpenAllocationAccountRequest,
+} from '@daml.js/da-marketplace/lib/Marketplace/Custody/Model';
+import { VerifiedIdentity } from '@daml.js/da-marketplace/lib/Marketplace/Regulator/Model';
+
+type OfferProps = {
+  offer?: CreateEvent<Offer>;
+};
+
+const ClearingOfferModal: React.FC<ServicePageProps<Service> & OfferProps> = ({
```
alexgraham-da, line 29: I'd like to add this functionality (the auto-requesting) to the notifications/request dialogs as well, but for now it will only show up when you accept a clearing offer from the Clearing Network page.

ui/src/pages/clearing/ClearingOfferModal.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const [clearingAccountName, setClearingAccountName] = useState('');
+  const [marginAccountName, setMarginAccountName] = useState('');
+  const [clearingProvider, setClearingProvider] = useState(!!offer ? offer.payload.provider : '');
+  const allocationAccountRules = useStreamQueries(AllocationAccountRule).contracts;
+  const allocationAccounts = allocationAccountRules
+    .filter(c => c.payload.nominee === clearingProvider)
+    .filter(c => c.payload.account.owner === party)
+    .map(c => c.payload.account);
+
+  const identities = useStreamQueries(VerifiedIdentity).contracts;
```
Sarahb-da, line 45: there is also the `useVerifiedParties` hook in `config.ts`

ui/src/pages/landing/Landing.tsx
```diff
@@ -257,11 +261,27 @@ const Landing = () => {
       .reduce((sum, deposit) => sum + +deposit.payload.asset.quantity, 0)
   );
 
+  useEffect(() => {
+    console.log('state change');
```
Sarahb-da, line 265: lingering logs here, not sure if u meant to leave them

Sarahb-da (approved):

ui/src/pages/landing/Landing.tsx
```diff
@@ -257,11 +261,27 @@ const Landing = () => {
...
+    const provider =
+      identities.find(i => i.payload.legalName === dialogState?.provider)?.payload.customer || '';
+    const filteredFields: Fields = Object.fromEntries(
+      Object.entries(filterFields).map(([k, v]) => [k, v(provider)])
+    );
```
alexmatson-da, line 270: What exactly does this `Object.fromEntries(Object.entries(...)....)` stuff do? I think I saw another similar thing down below too

alexmatson-da: I suppose to be more precise with my question, what function is `v`?

alexgraham-da: I could definitely name it better than `v`, but it's a function `(provider: Party) => Field` that's meant to run every time the provider changes to get a filtered version of the list items (although you could  technically do anything you want, so maybe `filterField` is misnamed as well)...so an example of what `v` might be is:
```typescript
    provider => {
      return {
        label,
        type: 'selection',
        items: assetSettlementRules
          .filter(
            ar => ar.payload.observers.map.has(provider) || ar.payload.account.provider == provider
          )
          .map(ar => ar.payload.account.id.label),
      };
```

alexmatson-da: gotcha, thanks!

alexmatson-da (approved):

ui/src/pages/landing/Landing.tsx
alexgraham-da: Might be clearer like this: 
```typescript
const filteredFields: Fields = _.mapValues(fieldFromProviders, createFieldFn => createFieldFn(provider));
```

alexmatson-da: yeah! that's a lot better actually

alexmatson-da: and +1 for lodash, there are probably more opportunities to use it more often... so many useful little utilities in there

alexgraham-da: for sure!

alexgraham-da: I changed it to use the lodash function and these types (which I was able to reuse in the notifications side of it):
```typescript
export type FieldCallback<T> = (arg: T) => Field;
export type FieldCallbacks<T> = Record<string, FieldCallback<T>>;
```

I'm glad you asked for clarification, this  all makes it much cleaner!

PR #354, opened 2021-05-26
Multi login through Quick Setup 

Sarahb-da:
- changes nav reliance from state to deep links to accommodate new tab log in
- adds loading to review page
- adds ability to log in on a new tab
- removes public party from list of parties deploying autoapprove trigger


ui/src/pages/QuickSetup/AddPartiesPage.tsx
```diff
@@ -29,6 +56,50 @@ const AddPartiesPage = (props: { localOperator: string; onComplete: () => void }
     </label>
   );
 
+  if (loadingStatus === LoadingStatus.CREATING_ADMIN_CONTRACTS) {
```
Sarahb-da, line 59: I moved the logic for loading admin contracts and creating verified identities to this file from the quick setup file because this loading state happens under the `/add-parties` url

ui/src/pages/QuickSetup/QuickSetup.tsx
```diff
@@ -168,264 +127,144 @@ const QuickSetup = () => {
...
+                    </NavLink>
+                  )}
+                </>
+              ))}
             </Menu>
           )}
-          {activePage}
-        </div>
 
-        {loadingStatus === LoadingStatus.CREATING_ADMIN_CONTRACTS && (
```
Sarahb-da, line 214: these now belong in `AddPartiesPage.tsx`

ui/src/context/UserContext.tsx
```diff
@@ -85,14 +85,17 @@ function useUserDispatch() {
...
-  credentials: Credentials
+  credentials: Credentials,
+  newTab?: boolean
 ) {
   const { party, token } = credentials;
 
   try {
+    if (newTab) {
+      window.open('/app#/quick-setup/log-in-parties');
+    }
```
Sarahb-da, line 96: opens a new tab that is actually the log in page, then logs in on current tab

ui/src/context/UserContext.tsx
```diff
@@ -85,14 +85,17 @@ function useUserDispatch() {
...
   history: History,
-  credentials: Credentials
+  credentials: Credentials,
+  newTab?: boolean
 ) {
   const { party, token } = credentials;
 
   try {
+    if (newTab) {
+      window.open('/app#/quick-setup/log-in-parties');
```
alexmatson-da, line 95: I think this should be `/#/app/quick-setup/log-in-parties`

ui/src/pages/QuickSetup/FinishPage.tsx
```diff
@@ -60,29 +60,46 @@ const LoginTileGrid = () => {
...
+    <>
+      <div className="setup-page finish">
+        <p className="dark details">cmd + click on a tile to Log In on a new tab </p>
+        <div className="log-in-tile-grid">
+          {identities.map(p => (
+            <div
+              className="log-in-tile"
+              key={p.payload.customer}
+              onClick={e => handleClick(e, p.payload.customer)}
+            >
```
alexmatson-da, line 72: Not all users will be on mac... windows/linux uses "ctrl + click" for opening a link in a new tab. 

I think a possibly better option is
```
<Link to={location => {  
       loginUser(dispatch, history, computeCredentials(party), newWindow); 
       return '/app' ;
} >
``` 

I don't know for sure if this function will work like this (the react router docs show it being used to modify the existing location https://reactrouter.com/web/api/Link/to-function ) 

but if it does, then the ctrl + click / cmd + click should work naturally across all platforms/browsers. 

Then we can just style the Link components like the tiles

Sarahb-da: I see, will give it a try!

alexmatson-da (approved):

PR #358, opened 2021-05-27
Service Requests from Quick Setup

Sarahb-da:
changes offer page to request page

the caveat is that now we cant see a list of requests as the user makes them (i tried a few different ways to mediate this but it started taking too much time. Can come back to it after the role contract is public)


brian-weir (approved):
Looks good!

alexmatson-da (approved):
lgtm

PR #359, opened 2021-05-27
Minor fixes

brian-weir:
Contains multiple fixes, mainly based of Eric's feedback :

Auction : i think you should have to choose the auction type (even if we only have one right now) so that it's clear to the user what they really chose
Auction : What visibility should a customer of an auction service have
Trading : An asset split across multiple AssetDeposits cannot be traded as the sum of their quantities
Trading : Matching engine over allocated on trade
Triggers : Settlement trigger should call the service, not the model template directly


alexmatson-da (approved):
looks good

Sarahb-da (approved):

PR #363, opened 2021-05-28
Material UI cleanup

alexmatson-da:


daml/Demo.daml
```diff
@@ -79,7 +79,7 @@ examples Providers{..} Assets{..} issuer = do
   (_, del)    <- originate "DEL-TSLA" "Delivery Example" $ delivery tsla.assetId 20.0
   (_, dvp)    <- originate "DVP-TSLA-USD" "DvP Example" $ dvp tsla.assetId 20.0 usd.assetId 10_000.0
   (_, fwd)    <- originate "FWD-TSLA-USD-20221022" "Forward Example" $ futureDvp tsla.assetId 20.0 usd.assetId 10_000.0 (date 2022 Oct 22)
-  (_, bnd)    <- originate "BOND-USD-4.875" "Forward Example" $ bond usd.assetId 1000000.0 0.04875 $ unrollDates 2021 2023 [Jan,Jul] 15
+  (_, bnd)    <- originate "BOND-USD-4.875" "Bond Example" $ bond usd.assetId 1000000.0 0.04875 $ unrollDates 2021 2023 [Jan,Jul] 15
```
alexmatson-da, line 82: The repeated value of `"Forward Example"` here actually resulted in a runtime React error, because the asset dropdowns/selections use these labels as the keys of the components being iterated out on. 🤔 

this was obviously a typo, but we may need to think about how to better support having multiple assets with the same name in the UI, _or_ enforce asset name uniqueness in the Daml models

ui/src/pages/QuickSetup/QuickSetup.tsx
```diff
@@ -120,30 +121,27 @@ const QuickSetup = withRouter((props: RouteComponentProps<{}>) => {
...
-
-        <div className="quick-setup-header">
-          <h1 className="logo-header">
-            <OpenMarketplaceLogo size="32" /> Daml Open Marketplace
-          </h1>
-          {activeMenuItem === MenuItems.LOG_IN ? <h2>Log In</h2> : <h2>Market Set-Up</h2>}
-        </div>
-
-        <div className="quick-setup-tile">
+      <Widget
```
alexmatson-da, line 124: I pulled out some of the QuickSetup styling into a new component, `Widget`, so I could reuse it for the error page! 

I think this a nice layout for pages "outside" the main app 

Screen Shot 2021-05-28 at 1 17 50 PM

alexgraham-da: nice 👍

daml/Demo.daml
alexgraham-da: `AssetDescription` is keyed on the `assetId` which is the label and the signatories from the `Issuance` service `[provider, customer]`...so there is name uniqueness per `Issuance` service. So at least for the keys, if you keyed the dropdowns as `${assetId.label}-${assetId.signatories}`, that should work.

alexgraham-da (approved):
looks good! RIP material, long live semantic!

PR #364, opened 2021-05-28
Role view request

alexgraham-da:
adds a role tile to the landing page where the party can view their roles/requests, and request new roles. also made it so `Matching` and `Settlement` services are automatically created when the `Trading` role is created.


ui/src/components/InputDialog/RoleDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+import { useLedger } from '@daml/react';
+
+import { InputDialog } from './InputDialog';
+
+import { Template } from '@daml/types';
+import { RoleRequestTemplates, RoleKind } from '../../context/RolesContext';
+
+interface RequestProps<T extends RoleRequestTemplates> {
+  fields: any;
```
alexmatson-da, line 11: Would the following work?
 
```suggestion
interface RequestProps<T extends RoleRequestTemplates, K extends { [key: string]: any }> {
  fields: Record<keyof K, Field>;
```

This is how `fields` is typed in InputDialog. Don't spend too much time on this if it's not an easy win, but it's always nice to not rely on `any` when we can

alexmatson-da: Hah, of course, even the suggestion relies on `any` a bit anyway :p

ui/src/components/InputDialog/RoleDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  open?: boolean;
+  params: T;
+  request: Template<T, undefined, string>;
+  role: RoleKind;
+  onChange: (state: any) => void;
```
alexmatson-da, line 16: If the `K` type parameter above works, then

```suggestion
  onChange: (state: K | null) => void;
```

ui/src/pages/landing/Landing.tsx
```diff
@@ -171,187 +143,22 @@ const ProfileSection: React.FC<{ name: string }> = ({ name }) => {
...
-    provider => {
-      return {
-        label,
-        type: 'selection',
-        items: allocationAccountRules
-          .filter(ar => ar.payload.nominee === provider)
-          .map(ar => ar.payload.account.id.label),
-      };
-    };
+  console.log(Object.values(roleRequests.values));
```
Sarahb-da, line 161: leftover log

ui/src/pages/landing/RoleRequestMenu.tsx
```diff
@@ -0,0 +1,177 @@
...
+    () =>
+      assetSettlementRules
+        .filter(c => c.payload.account.owner === party)
+        .map(c => c.payload.account),
+    [party, assetSettlementRules]
+  );
+
+  useEffect(() => {
+    const operator =
+      identities.find(i => i.payload.legalName === dialogState?.operator)?.payload.customer || ''; //?.payload.customer || '';
```
Sarahb-da, line 67: comment ^

Sarahb-da (approved):
I feel like a future state of this should be abstracting away service requests from the user entirely. And only depending on roles + relationships to automatically create service contracts. Or, bc this is the "service-based" version of the marketplace, maybe roles are what we need to abstract away from the UI.  I think now more than ever it is redundant / kinda confusing that we separately request roles and services.

ui/src/components/InputDialog/RoleDialog.tsx
alexgraham-da: lol yep. I tried the above, which I then had to add the same constraint to `RoleRequestDialog` below...which then gave me this error for `InputDialog`'s `defaultValue` field:
```
tsserver 2322] [E] Type '{ operator: string; }' is not assignable to type 'K'.
  '{ operator: string; }' is assignable to the constraint of type 'K', but 'K' could be instantiated with a different subtype of constraint '{ [key: string]: any; }'.
```

I based this off of `ServiceDialog.tsx`, this may be why it used `any`...happy to try something else though, I'm curious why it won't work like this.

alexgraham-da:
> I feel like a future state of this should be abstracting away service requests from the user entirely. And only depending on roles + relationships to automatically create service contracts. Or, bc this is the "service-based" version of the marketplace, maybe roles are what we need to abstract away from the UI. I think now more than ever it is redundant / kinda confusing that we separately request roles and services.

@Sarahb-da I definitely agree with this. In this case it is particularly confusing that "Role" is only for providing a service, when something like "Investor" sounds like a role but it's entirely made up of various service contracts as a customer.

Maybe something closer to what we had in the old marketplace..."I would like to trade", "I would like to list markets", etc which then does the appropriate role/service requesting for the user.

For now it might make it more clear to change "Roles" to something like "Services Provided".

ui/src/components/InputDialog/RoleDialog.tsx
alexmatson-da: Hm yep, off the top of my head, not sure how to work around the subtype constraint. Then this is good for now

alexmatson-da (approved):

PR #365, opened 2021-06-01
more granular page labeling

Sarahb-da:
Screen Shot 2021-06-01 at 10 52 15 AM


ui/src/pages/custody/Account.tsx
```diff
@@ -254,13 +254,10 @@ const AccountComponent: React.FC<RouteComponentProps & ServicePageProps<Service>
 
   return (
     <>
-      <BackButton />
+      <BackButton prevPage="Wallet" />
```
alexmatson-da, line 257: One thing to note, the label here and the actual page you get to when you go back might not always match up. If you're on the Manage page for example, and then go to `/app/custody/account` in the address bar, then the label will say "Back to Wallet" but will take you to the Manage page. 

Maybe add another optional prop with a path, and have the button go there, and fallback to `history.goBack()` if it is not given

Sarahb-da: Added an optional path!

alexmatson-da (approved):
nice nice!

PR #366, opened 2021-06-01
Auto deploy automation

Sarahb-da:
deploy triggers automatically based on the role selected


ui/src/pages/QuickSetup/SelectRolesPage.tsx
```diff
@@ -134,6 +152,19 @@ const DragAndDropRoles = () => {
     }
   }
 
+  async function handleDeployment(token: string, autoName: string) {
+    console.log(allTriggers);
+    const trigger = allTriggers.find(auto => auto.startsWith(autoName));
+    if (trigger) {
+      console.log(trigger);
```
alexmatson-da, line 159: console logs

alexmatson-da (approved):
lgmt

ui/src/pages/QuickSetup/SelectRolesPage.tsx
```diff
@@ -111,6 +122,9 @@ const DragAndDropRoles = () => {
           operatorServiceContract.contractId,
           provider
         );
+        if (isHubDeployment) {
+          handleDeployment(token, MarketplaceTrigger.MatchingEngine);
```
alexgraham-da, line 126: In the latest `main`, `OfferExchangeRole` will  now automatically setup the Matching and Settlement services. Since Matching is only needed with trading, you won't need to have the user select the matching service manually. Settlement is not trading specific, so that we should leave as selectable.

Also, is there any way to have the user select if they want the Matching Engine? If the user is planning to use Exberry, they would only deploy the Exberry Adapter and not the Matching Engine trigger. If there's no good way to have it work in the UI, the user can always undeploy the matching engine and switch it to the adapter in the setup page (or console).

alexgraham-da (approved):
just the one comment, but looks good otherwise

PR #374, opened 2021-06-08
Remove added roles

alexmatson-da:


daml/Marketplace/Clearing/Role.daml
```diff
@@ -111,3 +111,20 @@ template Request
...
+template RequestRoleTermination
+  with
+    operator : Party
+    provider : Party
+    roleCid  : ContractId Role
+  where
+    signatory provider
+
+    controller operator can
+      TerminateAccept: ()
```
alexmatson-da, line 124: I had to give these awkward names to avoid a name conflict with the `Accept` and `Decline` in the `Offer` template. Why does Daml not scope these?!

daml/Marketplace/Clearing/Role.daml
```diff
@@ -111,3 +111,20 @@ template Request
       Reject : ()
         do
           return ()
+
+template RequestRoleTermination
```
alexgraham-da, line 115: Is using request contracts a design decision to require confirmation by the operator to terminate a role? If not, why not just add a consuming `Terminate` choice on the role contract with the `provider` as the `controller`? That should work, unless I'm missing something

ui/src/pages/QuickSetup/SelectRolesPage.tsx
```diff
@@ -185,10 +194,60 @@ const DragAndDropRoles = () => {
...
+
+  return (
+    <div
+      className="service-label"
+      onMouseEnter={() => setHovering(true)}
+      onMouseOut={() => setHovering(false)}
+    >
+      {hovering ? (
+        <Label basic onClick={() => terminateRole(role, ledger)}>
+          {role.roleKind} <Icon name="close" />
```
Sarahb-da, line 214: Ive been trying to only use our custom icons because they are a bit different than the ones provided by semantic (and we have them anyway) if possible..

Sarahb-da: Also, why not just use `:hover` and `display: none` in scss instead?

daml/Marketplace/Clearing/Role.daml
alexmatson-da: The Roles contracts have operator and provider as signatories

Sarahb-da (approved):
lgtm!

daml/Marketplace/Clearing/Role.daml
brian-weir: I feel your pain, this gets very annoying very quickly... I used to use <template_name>_<choice_name> as a pattern to get around it but this also isn't great.

brian-weir (approved):
lgtm! 👍

daml/Marketplace/Clearing/Role.daml
alexgraham-da: Right - but having a choice on the contract with only the `provider` as the controller would give the controller party authorization to archive the contract (provided it is within that choice), without explicit permission during the exercise from the other signatory

alexmatson-da: I'll give it a try - I like the idea of having a terminate choice better than the request templates

daml/Marketplace/Clearing/Role.daml
alexmatson-da: Fortunately @alexgraham-da 's suggestion works, so the terminate choices can live on the Role contract itself without needing a separate contract!

PR #380, opened 2021-06-11
Clearing model fixes

brian-weir:


daml/Tests/Distribution/Auction.daml
```diff
@@ -260,8 +260,11 @@ assertAssetDeposits auction issuer originalDeposits price quantity bid = do
   quotedDeposit <- queryFilter @AssetDeposit bid.customer (\deposit -> deposit.account == bid.tradingAccount && deposit.asset.id == auction.quotedAssetId)
   case quotedDeposit of
     ([(_, x)]) -> do
-      let originalDeposit = filter (\deposit -> deposit.account.owner == x.account.owner) originalDeposits
-      assertMsg "Investor/Bidder split deposit contains incorrect quantity" (x.asset.quantity == (bid.details.quantity * price))
+      let originalDeposit = head $ filter (\deposit -> deposit.account.owner == x.account.owner) originalDeposits
+      let newQuantity = case bid.status of
+            (Bidding.PartialAllocation _ quantity) -> quantity
+            _                                      -> bid.details.quantity
+      assertMsg "Investor/Bidder split deposit contains incorrect quantity" (x.asset.quantity == (originalDeposit.asset.quantity - (newQuantity * price)))
```
alexgraham-da, line 267: this previously wasn't getting called due to the asset id's for USD having different signatories, now that it is getting called, this fixes the assert.

ui/src/pages/clearing/MTMCalculationModal.tsx
```diff
@@ -39,6 +39,12 @@ const MTMCalculationModal: React.FC<ServicePageProps<Service> & MTMProps> = ({
     }
   }, [member]);
 
+  useEffect(() => {
+    if (!currencyId && assets.length > 0) {
+      setCurrencyId(assets.find(c => c.payload.assetId.label === 'USD')?.payload.assetId);
+    }
+  }, [currencyId, assets]);
```
alexmatson-da, line 46: Does this mean the UI currently only does MtM with USD, and no other asset/currency?

brian-weir: @alexgraham-da can elaborate on this more than I can but my current understanding is that it works only with `USD` atm.

alexgraham-da: yes, that is correct. at some point this will be expanded, but that was the minimum requirement for the baymarkets demo, so we left it at that for the time being.

Sarahb-da (approved):
looks good to me!

alexmatson-da (approved):

PR #384, opened 2021-06-15
add requests page and a few other fixes

Sarahb-da:
sidebar hierarchy re-org stuff cont...


ui/src/pages/notifications/NotificationComponents.tsx
```diff
@@ -174,6 +174,7 @@ export function RequestNotification<T extends Fields>({
               className="ghost"
               content="Approve"
               type="submit"
+              disabled={!lookupFields}
```
Sarahb-da, line 177: There should also be a warning here, when fields are missing that block the user from accepting a request

brian-weir (approved):
lgtm !!

PR #385, opened 2021-06-15
Flatten wallet information 

Sarahb-da:
Instead of a Holdings table followed by an Account table, we now have holdings listed under their designated accounts and a quick deposit tile like the previous marketplace had. This kind of refactor is also working toward flattening the UX where we can i.e. reducing nested pages 

Screen Shot 2021-06-15 at 2 31 04 PM


alexgraham-da (approved):
looks good!

PR #387, opened 2021-06-23
Add exchange fees

alexgraham-da:
Adds Exchange Fees via an `Optional` `ContractId AssetDeposit` field in the order details and a customizable `FeeSchedule` contract that the Exchange can set. Orders that do not include the correct fee will be rejected by the MatchingEngine or Exchange Adapter (this is also enforced in the UI). When the order is matched and confirmed, the fees are moved into an account held by the Exchange. Fees are prorated when an order is partially filled.

Screen Shot 2021-06-22 at 8 34 32 PM

### DIFF
--- a/daml/Marketplace/Trading/Confirmation/Model.daml
+++ b/daml/Marketplace/Trading/Confirmation/Model.daml
@@ -7,10 +7,8 @@ import DA.Finance.Types (Account(..))
 import DA.Finance.Utils (fetchAndArchive)
 import DA.Set (Set, insert)
 import Marketplace.Settlement.Model qualified as Settlement
-import qualified Marketplace.Rule.AllocationAccount as AllocationAccountRule (T, Transfer(..))
 import Marketplace.Trading.Model qualified as Order
-import DA.Optional.Total (whenSome)
-import DA.Functor (void)
+import DA.Optional.Total (optionalToList)
 
 -- TODO: Try to turn this into a service
 
@@ -82,7 +80,7 @@ template Confirmation
           -- TODO: Check that price matches limit order price, quantity is less than order qty, rounding, etc
           -- paymentCurrencyId <- (.quotedAssetId) . snd <$> fetchByKey @Listing (operator, provider, buyOrder.details.symbol)
 
-          [payment, delivery] <- forA [(buyOrder, buyer, seller), (sellOrder, seller, buyer)] (\(order, party, counterparty) -> do
+          details <- concat <$> forA [(buyOrder, buyer, seller), (sellOrder, seller, buyer)] (\(order, party, counterparty) -> do
             let
               isPartialFill = execution.quantity < order.remainingQuantity
               executions = execution :: order.executions
@@ -93,38 +91,40 @@ template Confirmation
               getAllocationAccount = getAccount accounts (.allocationAccount)
               getTradingAccount    = getAccount accounts (.tradingAccount)
 
-            let transferToFeeAccount : ContractId AssetDeposit -> Update (ContractId AssetDeposit)
-                transferToFeeAccount depositCid = do
+            let createFeeDetails : ContractId AssetDeposit -> Update Settlement.SettlementDetails
+                createFeeDetails depositCid = do
                   feeAccount <- (.feeAccount) . snd <$> fetchByKey @Order.FeeSchedule (operator, provider)
-                  exerciseByKey @AllocationAccountRule.T (getAllocationAccount party).id
-                      AllocationAccountRule.Transfer with transferTo = feeAccount; ..
+                  pure $ Settlement.SettlementDetails with depositCid; senderAccount = getAllocationAccount party; receiverAccount = feeAccount
 
             let (Order.Collateral depositCid) = order.collateral
-            depositCid <- if isPartialFill
+            (depositCid, optFeeDetails) <- if isPartialFill
               then do
-                newOptExchangeFee <- case order.details.optExchangeFee of
-                  None               -> return None
+                (newOptExchangeFee, feeDetails) <- case order.details.optExchangeFee of
+                  None               -> return (None, None)
                   Some feeDepositCid -> do
                     feeQuantity <- (.asset.quantity) <$> fetch feeDepositCid
                     let feeToTake = roundBankers 2 $ feeQuantity * (execution.quantity / order.remainingQuantity)
                     [feeDepositCid, remainingFeeDepositCid] <- exercise feeDepositCid AssetDeposit_Split with quantities = [feeToTake]
-                    transferToFeeAccount feeDepositCid
-                    return $ Some remainingFeeDepositCid
+                    feeDetails <- createFeeDetails feeDepositCid
+                    return $ (Some remainingFeeDepositCid, Some feeDetails)
 
                 (depositCid :: depositRemainingCid :: _) <- exercise depositCid AssetDeposit_Split with ..
                 create order with
                   executions, remainingQuantity
                   collateral = Order.Collateral depositRemainingCid, status = Order.PartiallyExecuted
                   details = order.details with optExchangeFee = newOptExchangeFee
-                pure depositCid
+                pure (depositCid, feeDetails)
               else do
-                whenSome order.details.optExchangeFee (void . transferToFeeAccount)
+                optFeeDetails <- case order.details.optExchangeFee of
+                  Some feeCid -> Some <$> createFeeDetails feeCid
+                  _           -> return None
                 create order with
                   executions, remainingQuantity
                   status = Order.FullyExecuted, details = order.details with optExchangeFee = None
-                pure depositCid
+                pure (depositCid, optFeeDetails)
 
-            pure Settlement.SettlementDetails with depositCid; senderAccount = getAllocationAccount party; receiverAccount = getTradingAccount counterparty
+            let tradeDetails = Settlement.SettlementDetails with depositCid; senderAccount = getAllocationAccount party; receiverAccount = getTradingAccount counterparty
+            pure $ optionalToList optFeeDetails <> [tradeDetails]
             )
 
           create Settlement.SettlementInstruction with ..
--- a/daml/Marketplace/Trading/Confirmation/Model.daml
+++ b/daml/Marketplace/Trading/Confirmation/Model.daml
@@ -2,13 +2,15 @@ module Marketplace.Trading.Confirmation.Model where
 
 import DA.List (head)
 import DA.Foldable (forA_)
-import DA.Finance.Asset (AssetDeposit_Split(..))
+import DA.Finance.Asset (AssetDeposit, AssetDeposit_Split(..))
 import DA.Finance.Types (Account(..))
 import DA.Finance.Utils (fetchAndArchive)
 import DA.Set (Set, insert)
 import Marketplace.Settlement.Model qualified as Settlement
--- import Marketplace.Trading.Listing (Listing)
+import qualified Marketplace.Rule.AllocationAccount as AllocationAccountRule (T, Transfer(..))
 import Marketplace.Trading.Model qualified as Order
+import DA.Optional.Total (whenSome)
+import DA.Functor (void)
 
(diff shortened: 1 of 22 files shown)


daml/Marketplace/Trading/Model.daml
```diff
@@ -93,3 +95,32 @@ template Order
     key (provider, details.id.label) : (Party, Text)
     maintainer key._1
 
+data Fee = Fee with
+    amount : Decimal
+    currency : Id
+    timeInEffect : Time
+  deriving (Show, Eq)
```
brian-weir, line 102: This is the `Fee` per trade ?

daml/Marketplace/Trading/Confirmation/Model.daml
```diff
@@ -91,14 +93,35 @@ template Confirmation
...
+                  executions, remainingQuantity
+                  collateral = Order.Collateral depositRemainingCid, status = Order.PartiallyExecuted
+                  details = order.details with optExchangeFee = newOptExchangeFee
                 pure depositCid
               else do
-                create order with executions, remainingQuantity, status = Order.FullyExecuted
+                whenSome order.details.optExchangeFee (void . transferToFeeAccount)
+                create order with
+                  executions, remainingQuantity
+                  status = Order.FullyExecuted, details = order.details with optExchangeFee = None
```
brian-weir, line 124: So we take the `Fees` on every execution received for an order ? Ie, we take them _in real time_ regardless if the execution has settled or not between participants ? If so, was that the requirement from Eric ? 

I'm just wondering if the payment of the fee should be part of the settlement process. Ie, we would exchange the assets between the holders and pay the exchange the fees as part of one atomic transaction.

exberry_adapter/bot/exberry_adapter_bot.py
```diff
@@ -99,6 +100,21 @@ def handle_order_request(event):
...
+            fee_amount = fee_deposit.cdata['asset']['quantity']
+
+        fee_schedules = client.find_active(MARKETPLACE.FeeSchedule, {'provider': exchange_party})
+        for (_,schedule) in fee_schedules.items():
+            if schedule['currentFee']['amount'] > fee_amount:
+                return [exercise(event.cid,
+                            'RejectRequest', {
+                                'errorCode': '790',
+                                'errorMessage': 'Fee does not match exchange fee'}
+                        )]
```
brian-weir, line 116: I'm not quite sure how I feel about having this business logic (or check) here in the Exberry adapter.

Could we not have this logic sit in Daml when a user creates an order ? Is the issue that the `customer` cannot see the `FeeSchedule` contract as it's not an `observer` on that contract nor has the signature from the `provider ?

ui/src/pages/network/Fees.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const accountRule = assetSettlementRules.find(ar => ar.payload.account.id.label === accountLabel);
+
+  const canRequest = !feeSchedule
+    ? !!assetLabel && !!asset && !!accountLabel && !!account && !!quantity
+    : !!quantity;
+
+  if (!role)
+    return (
+      <div>
+        <h2>Party "{party}" can not setup a Fee Schedule.</h2>
```
Sarahb-da, line 56: maybe worth adding - "To set up a fee schedule, add the exchange role and create an asset.. "

ui/src/pages/network/Fees.tsx
```diff
@@ -0,0 +1,140 @@
...
+      </div>
+    );
+
+  const createFeeSchedule = async () => {
+    if (!feeSchedule) {
+      if (!asset || !account) return;
```
Sarahb-da, line 62: display error message here?

Sarahb-da: and possibly elsewhere in this `createFeeSchedule`

ui/src/pages/network/Fees.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+            <Form.Input
+              placeholder={!!feeSchedule ? feeSchedule.payload.currentFee.amount : quantity}
+              required
+              fluid
+              number
+              label="Fee Amount"
+              value={quantity}
+              onChange={e => setQuantity(e.currentTarget.value)}
+            />
```
Sarahb-da, line 116: Is there a common percentage/suggested value or something we can use here? I am just thinking about what eric said about raw input fields. Would be cool if you could specify a percentage or an amount. Just a thought.

ui/src/pages/network/Fees.tsx
```diff
@@ -0,0 +1,140 @@
...
+      await ledger.exercise(FeeSchedule.UpdateFeeSchedule, feeSchedule.contractId, {
+        amount: quantity,
+        currency: asset.payload.assetId,
+      });
+    }
+  };
+
+  return (
+    <div className="page-section-row">
+      <Tile className="inline" header="Setup Fees">
```
Sarahb-da, line 79: you could add `loading=schedulesLoading` now that you can pass that prop to Tile

ui/src/pages/network/Fees.tsx
```diff
@@ -0,0 +1,140 @@
...
+              onChange={e => setQuantity(e.currentTarget.value)}
+            />
+
+            <Button type="submit" disabled={!canRequest} className="ghost">
+              {!!feeSchedule ? 'Update Fee Schedule' : 'Create Fee Schedule'}
+            </Button>
+          </FormErrorHandled>
+        )}
+      </Tile>
+      <Tile className="inline">
```
Sarahb-da, line 124: similarly you could add `loading= assetSettlementRulesLoading`

ui/src/pages/network/Trading.tsx
```diff
@@ -15,39 +18,44 @@ export const TradingServiceTable: React.FC<Props> = ({ services }) => {
   const { getName } = usePartyName(party);
   const ledger = useLedger();
 
+  const role = useStreamQueries(ExchangeRole).contracts.find(rl => rl.payload.provider === party);
```
Sarahb-da, line 21: idea for future PR - `useRole` hooks

ui/src/pages/network/Fees.tsx
Sarahb-da: oh wait I just realized line 29 in `trading.tsx` ensures this wont happen

ui/src/pages/trading/Market.tsx
```diff
@@ -245,22 +255,38 @@ export const Market: React.FC<ServicePageProps<Service> & Props> = ({
...
     if (listing.payload.listingType.tag === 'Collateralized') {
       const depositCid = isBuy
         ? await getAsset(quotedAssets, price * quantity)
         : await getAsset(tradedAssets, quantity);
-
       if (!depositCid) return;
+      let optExchangeFee = null;
+      if (feeAmount > 0) {
+        const feeAssets = await ledger
+          .query(AssetDeposit)
```
Sarahb-da, line 269: should we gather assetDeposits independently with `useStreamQueries`? I hvnt seen `.query` used in this app yet, but your call.

Sarahb-da (commented):
just some front end stuff.

daml/Marketplace/Trading/Model.daml
alexgraham-da: This is the fee per order. As the order is filled, the fee is split and taken by the Exchange until the order is completely filled.

daml/Marketplace/Trading/Confirmation/Model.daml
alexgraham-da: The payment of the fee is part of the settlement process - this is just locking the fee in the `AllocationAccount` to prevent it from being used elsewhere (similar to orders).

exberry_adapter/bot/exberry_adapter_bot.py
alexgraham-da: I would have preferred this to not be in the adapter as well, but you are correct that the issue is the `customer` cannot see the contract, so therefore would not be able to look it up in `CreateOrder` to perform this check. Do you have any other ideas on how to do this? I couldn't think of a way of doing it without multi-party submits or a trigger adding all Trading customers to the `FeeSchedule` `observers` list.

ui/src/pages/network/Fees.tsx
alexgraham-da: I might as well make the error clearer anyway in case this component is used elsewhere

ui/src/pages/trading/Market.tsx
alexgraham-da: The issue here is that the asset deposits from the `useStreamQuery` may be changed in the preceding `getAsset` (if the currencies were the same) and the stream wasn't getting updated quickly enough before calling `getAsset` again here, performing a manual query fixed this.

ui/src/pages/network/Fees.tsx
alexgraham-da: I'm not sure what you mean by percentage, this is for a flat fee (although a percentage fee would be interesting)...maybe I will just default it to `0.0` so it at least there is something there.

ui/src/pages/network/Fees.tsx
alexgraham-da: ah I did have that there but it had still stopped loading before the contract actually showed up. I'll add it back though as it could still be useful if that part of the loading is being slow (and is definitely better than just not showing the content at all). 👍

daml/Marketplace/Trading/Confirmation/Model.daml
brian-weir: I thought these two lines did an actual transfer of the fee to the exchange or am I mistaken :

`transferToFeeAccount feeDepositCid`
`return $ Some remainingFeeDepositCid`

and 

`whenSome order.details.optExchangeFee (void . transferToFeeAccount)`

ui/src/pages/network/Fees.tsx
Sarahb-da: 0.0 sounds good - I was thinking the fee was a percentage of some traded quantity. Or it could be optionally represented that way. But since it is a flat fee I guess that wouldnt make much sense!

daml/Marketplace/Trading/Confirmation/Model.daml
alexgraham-da: Ah, I apologize, I thought this was in `Trading.Service`! Is the idea to move this to `SettlementInstruction` then?

alexgraham-da: Moving the transaction to `SettlementInstruction` worked well, see: https://github.com/digital-asset/da-marketplace/pull/387/commits/389532d304196c899a525b39569c210f51c8ea4a

ui/src/pages/network/Trading.tsx
alexgraham-da: yeah that would be great! I just tried adding it to `RolesContext` quickly and I was getting type errors (possibly because mixing Roles and Services?) so I'll leave it like this for now

daml/Marketplace/Trading/Confirmation/Model.daml
brian-weir: Awesome, thanks @alexgraham-da 👍

Sarahb-da (approved):
lgtm!

exberry_adapter/bot/exberry_adapter_bot.py
GeorgSchneider: I'd say a fee schedule contract exists for every customer. You could imagine different fee tiers depending on volume traded, so then you need to be able to associate a user with a specific fee schedule. Also, it would allow to set custom fees for some customers. The fee schedule would be something that is agreed to and created upon request or grant of the service. That would then also solve this problem as it would be visible to the user.

alexgraham-da: @georg-da I agree that's the way to go - I discussed this with Eric initially and he requested we start with a single Fee Schedule per exchange and open it up to per customer at a later time. I could also see fees being different per listing.

PR #390, opened 2021-06-24
remove setup

Sarahb-da:
Forgot to remove this from sidebar and a few small ui corrections

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -37,7 +37,6 @@ import {
   ExchangeIcon,
   MegaphoneIcon,
   OrdersIcon,
-  ToolIcon,
   WalletIcon,
   IconMailLetter,
 } from './icons/icons';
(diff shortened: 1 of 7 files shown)


alexmatson-da (approved):

ui/src/App.tsx
```diff
@@ -330,36 +328,9 @@ const AppComponent = () => {
...
-    ],
-    additionalRoutes: [
-      {
-        path: paths.app.markets.order + '/:contractId',
-        render: () => <TradingOrder listings={listings} />,
-      },
       {
-        path: paths.app.markets.offer,
-        render: () => <Offer service={ServiceKind.TRADING} />,
+        path: paths.app.setup.identity,
```
alexmatson-da, line 332: If we're removing the setup page, does it still make sense to have a `/app/setup/identity` path?

Sarahb-da: yeah good point ill remove that

Sarahb-da: bye bye setup

PR #393, opened 2021-06-30
Bump contingent-claims deps

lucianojoublanc-da:
Includes new `expiry` and `payoffs` functions


daml/Marketplace/Custody/Model.daml
```diff
@@ -124,9 +123,11 @@ template LifecycleRequest
               spot label dt = pure $ fromSomeNote
                 ("Lifecycle: Underlying data '" <> label <> "' missing on " <> show dt)
                 (Map.lookup label fixings >>= Map.lookup dt)
-              election : Claim Observation Date Id -> Claim Observation Date Id -> Claim Observation Date Id
-              election _ _ = deserialize choice -- will throw if the claims don't match
-          Lifecycle.Result{remaining, pending} <- Lifecycle.lifecycle spot election (deserialize description.claims) t
+              election = (True, deserialize choice)
```
lucianojoublanc-da, line 126: With the updates to `contingent-claims`, the exercising/lifecycling have been split into separate functions.

I've merged this with the aim of minimising the delta, which will also keep the functionality and API the same, but we may want to re-visit our approach to lifecycling in future; currently it has to be initiated by the customer, but we could now. E.g. create a trigger run by the custodian that attempts to lifecycle stuff automatically. This makes sense from a practical point of view (in the real world you don't need to request coupon or dividend payments - they happen automatically).

lucianojoublanc-da:
@alexgraham-da please checkout https://github.com/digital-asset/contingent-claims/blob/bddda6859f26a0cb6fe0fe0272a5e0db15c1487c/test/daml/Test/Util.daml#L71 for examples how to use the new `expiry` and `payoffs` functions. Implementing a `multiplier`, `underlying` and `maturity` should be trivial now.

ui/src/pages/origination/NewBinaryOption.tsx
```diff
@@ -55,29 +55,27 @@ const NewBinaryOptionComponent = ({ history }: RouteComponentProps) => {
...
-    value: { _1: obsToday, _2: obsExpiry },
+  const ineqEuropean: Inequality<DamlDate, Id> = {
+    tag: 'TimeGte',
+    value: parseDate(expiry),
   };
-  const obsStrike: Observation<DamlDate, Id> = { tag: 'DecimalConst', value: strike };
-  const obsSpot: Observation<DamlDate, Id> = { tag: 'DecimalObs', value: underlying };
-  const obsPayoff: Observation<DamlDate, Id> = {
-    tag: 'DecimalLte',
+  const obsStrike: Observation<DamlDate, string> = { tag: 'Const', value: { value: strike } };
```
lucianojoublanc-da, line 62: The type here and below is actually `DamlDate, Decimal`.

Is it ok to leave it as `string`, or is there a better way to type this?

ui/src/components/Claims/util.ts
```diff
@@ -85,24 +51,48 @@ const transformObservation = (obs: Observation<Date, boolean>, linkText: string)
...
+  inequality: Inequality<Date, string>,
+  linkText: string
+): any => {
+  switch (inequality.tag) {
+    case 'TimeGte':
+      return {
+        ...inequality,
+        linkText,
+        type: 'Observation',
+        children: [], // todo
```
brian-weir, line 71: Are you planning on implementing this in another PR in the future or as part of this PR ?

ui/src/pages/origination/NewBinaryOption.tsx
brian-weir: `DamlDate` is a `string` in Typescript :

`import { Date as DamlDate } from '@daml/types';`

From the imported library "@daml/types" -> `export declare type Date = string;`

The Daml type `Decimal` is also a `string`, we can see in the "@daml/types" library that this is set as :

`export declare type Numeric = string;`
`export declare type Decimal = Numeric;`

The explanation in the library for typing `Decimal`s as `string`s in Typescript is :

> The counterpart of Daml's `Numeric` type. We represent `Numeric`s as string in order to avoid a loss of precision. The string must match the regular expression `-?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+-]?\d+)?`.

**TR;DR** - you're safe as the daml library is typing both of those Daml types as `string`s in Typescript although you could set this as `Observation<DamlDate, Decimal>` and import `Decimal` from the daml typescript library to be more explicit.

ui/src/components/Claims/util.ts
lucianojoublanc-da: Thanks, yes I missed this. Having some trouble with this though; I can't get the time parameter to show even if when this is compiling. Dunno if @georg-da D3 expert you can give me a hand with this.

GeorgSchneider (approved):

PR #395, opened 2021-06-30
react flow for review

Sarahb-da:
Screen Shot 2021-07-06 at 11 08 45 AM


package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "dagre": "^0.8.5"
```
alexmatson-da: For some reason, this `package.json` (and associated `package-lock.json`) seems to have wound up in the root directory, instead of the `ui/` directory.

Also, we should stick to using yarn -- so to fix, you'll need to delete these files, go into the `ui` dir, and run `yarn add dagre`, and afterwards `yarn install` to be safe. Then commit `yarn.lock` and things should start working, I think!

ui2/yarn.lock
```diff
@@ -0,0 +1,41 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
alexmatson-da, line 1: Oof, this file is in `ui2/` which is a directory that seems to have come back from the dead here! Not sure why that happened 😅 

Try a `git rm -rf ui2/` if you're actually seeing it locally

Sarahb-da: oh man, okay not sure how this happened

PR #399, opened 2021-07-07
Fix admin setup

Sarahb-da:
Fixes bug where you need to refresh during the `Confirming Admin Credentials` quick setup


ui/src/pages/QuickSetup/AddPartiesPage.tsx
```diff
@@ -260,40 +273,51 @@ const AdminLedger = (props: { adminCredentials: Credentials; onComplete: () => v
...
     }
+
     if (operatorService.length === 0) {
       createOperatorService();
     } else if (regulatorRoles.length === 0) {
       createRegulatorRole();
     } else {
       offerRegulatorServices();
+      deployAllTriggers();
+      return onComplete();
```
Sarahb-da, line 320: I think it is better to deploy triggers here, so that there is no race condition between creating and accepting regulator offers - and the UI can reliably check if all parties have been offered the role

Sarahb-da: Also, here the party list is fully loaded where as in quick setup that was not a guarantee and sometimes required a refresh

alexmatson-da (approved):

alexgraham-da (approved):
looks good to me 👍

PR #403, opened 2021-07-12
switch account selection to be inline on review page

alexgraham-da:
Changes the Quick Setup account selection (with automatic account creation) to be inline fields as a list of required accounts instead of modals

Screen Shot 2021-07-09 at 1 45 35 PM


Sarahb-da (approved):
👍 👍

alexmatson-da (approved):

PR #413, opened 2021-08-19
v1: updates for hub react

alexmatson-da:


ui/src/Credentials.ts
```diff
@@ -27,11 +27,11 @@ function isCredentials(credentials: any): credentials is Credentials {
 
 const CREDENTIALS_STORAGE_KEY = 'credentials';
 
-export function storeCredentials(credentials?: Credentials): void {
+export function storeCredentials(credentials?: PartyToken): void {
```
Sarahb-da, line 30: since we now have `Credentials` and `PartyToken` should we rename the arg `partyToken?: PartyToken `?

ui/src/Parties.ts
```diff
@@ -1,30 +1,28 @@
...
 
 const PARTIES_STORAGE_KEY = 'imported_parties';
 
-export function storeParties(parties: PartyDetails[]): void {
+export function storeParties(parties: PartyToken[]): void {
     localStorage.setItem(PARTIES_STORAGE_KEY, JSON.stringify(parties));
 }
 
-export function retrieveParties(validateParties: boolean = true): PartyDetails[] | undefined {
+export function retrieveParties(publicParty: string, validateParties: boolean = true): PartyToken[] | undefined {
```
Sarahb-da, line 9: `retrievePartyTokens`? Just for clarity bc the object changed?

ui/src/components/common/common.ts
```diff
@@ -14,24 +14,31 @@ type Result = {
...
     useEffect(() => {
-        if (deploymentMode === DeploymentMode.PROD_DABL) {
-            if (error && !loading) {
-                console.error(`Error fetching DABL parties: ${error}`);
-            }
+        if (!publicParty) {
+            setLoading(true);
+        } else {
+            setLoading(false);
+        }
```
Sarahb-da, line 29: or `setLoading(!publicParty)`

Sarahb-da (commented):
a few minor comments - the one thing I'll add is that now some variables called `credentials` are no longer `Credentials` but `PartyToken`s. Maybe we change var names accordingly to avoid confusion

alexgraham-da (approved):
nice, lgtm!

ui/src/Parties.ts
alexmatson-da: Hmm... `PartyToken` contains more than just the token, but also all the info about the party too (that was embedded in the token)

ui/src/Credentials.ts
alexmatson-da: Actually I moved this back to `Credentials`. I ought to be more careful because `PartyToken` is Hub-specific, so `Credentials` is actually more general (i.e. for local dev) 

i revised the PR to try to make sure `PartyToken` stuff is only contained within parts of the code that are only supposed to run on Daml Hub

PR #416, opened 2021-08-24
fix adapter startup and add test

alexgraham-da:


exberry_adapter/test/daml/Tests/Matching.daml
```diff
@@ -0,0 +1,98 @@
+module Tests.Matching where
```
alexgraham-da, line 1: I believe there used to be an `integrationTesting` model in this repo (I'm not sure when or why it was removed). This is similar, but runs the full test automatically. The test is basic but could be extended, possibly for a larger volume stress test.

daml.yaml
```diff
@@ -5,6 +5,8 @@ init-script: Demo:demo
 parties:
   - Alice
   - Bob
+  - Dana
+  - Carl
```
alexmatson-da, line 9: Do we use these parties in our tests? They are newly introduced in this PR, and the only other place I see them referenced is in `getExistingParties`. Can they be dropped?

alexgraham-da: They are used elsewhere in `Tests.Clearing.Setup` and I think `Distribution.Auction` (although I see now it's `Charlie` not `Carl`, so I will change). They aren't used in every test, but they are allocated in all tests via `allocateParties`. Not sure if it really helps or hurts having them all here, I believe the only thing it does is populate the navigator dropdown.

alexmatson-da: Yep makes sense, it doesn't hurt to have them if they are used in tests

alexmatson-da (approved):

Sarahb-da (approved):

exberry_adapter/test/daml/Tests/Matching.daml
brian-weir: It was removed as the code of the `integrationTesting` went out of sync very quickly and it was a very manual task to execute the testing in that you had to modify the code for almost each test followed by calling the various functions available in the correct order. The purpose of the `integrationTesting` was to test our daml models with exberry before the UI was finished.

brian-weir (approved):

PR #419, opened 2021-08-31
Locking upgrade and removal of asset rules

brian-weir:


alexmatson-da (commented):
I know this is still a draft, but I took a look at the work (focused more on the UI side) and it's looking good to me so far! 

I'll review again when officially marked ready

ui/src/pages/trading/Market.tsx
```diff
@@ -186,6 +197,9 @@ export const Market: React.FC<ServicePageProps<Service> & Props> = ({
     setPercentage(0.0);
     setTotal(0.0);
     setExpiryDate(0);
+    setAccountLabel(
+      custodyServices.length === 1 ? custodyServices[0].payload.account.id.label : ''
+    );
```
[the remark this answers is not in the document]
brian-weir: This has been removed

ui/src/pages/trading/Market.tsx
```diff
@@ -63,6 +67,10 @@ export const Market: React.FC<ServicePageProps<Service> & Props> = ({
   const [timeInForce, setTimeInForce] = useState<TimeInForces>('GTC');
   const [expiryDate, setExpiryDate] = useState(0);
 
+  const [accountLabel, setAccountLabel] = useState(
+    custodyServices.length === 1 ? custodyServices[0].payload.account.id.label : ''
+  );
```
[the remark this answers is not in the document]
brian-weir: Also removed

daml/Marketplace/Custody/Service.daml
```diff
@@ -190,4 +116,11 @@ template Request
...
       controller operator, provider
       do
+        let account = createAccount provider customer
         createOrLookup Service with ..
+
+createAccount : Party -> Party -> Account
+createAccount provider customer = do
+  let id = Id with signatories = fromList [customer, provider]; label = partyToText customer <> "@" <> partyToText provider; version = 0
+      owner = customer
+  Account with id; owner; provider
```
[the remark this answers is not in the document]
brian-weir: Fixed on the latest commit

ui/src/pages/trading/Order.tsx
```diff
@@ -145,7 +153,7 @@ export const TradingOrder: React.FC<Props> = ({ listings, services }: Props) =>
                 </Table.Row>
                 {order.payload.details.timeInForce.tag === 'GTD' && (
                   <>
-                    <Table.Row key={10}>
+                    <Table.Row key={11}>
```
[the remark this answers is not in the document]
brian-weir: I'll fix this in the next PR.

daml/Marketplace/Trading/Service.daml
```diff
@@ -53,30 +42,33 @@ template Service
...
             (Order.Collateral depositCid) -> do
-                details <- detailsWithDepositedFees
-                collateral <- Order.Collateral <$> exerciseByKey @AllocationAccountRule.T allocationAccount.id AllocationAccountRule.Deposit with ..
-                createOrderRequestCid <- create CreateOrderRequest with ..
-                orderCid <- create Order.Order with ..
-
-                return $ Right (orderCid, createOrderRequestCid, collateral)
+                lookupByKey @Custody.Service (operator, details.receivableAccount.provider, customer) >>= \case -- TODO: BDW - review this test
+                  None -> do
+                    let reason = Error with code = "3003"; message = "Customer does not have a custodial relationship with the custodian of the quoted asset"
```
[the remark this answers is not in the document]
brian-weir: Nope, we currently haven't documented them as there's only 3 or 4 error codes that we generate in the application. With `Exceptions` entering Daml, we should review the code base to see if we should utilise this new feature.

daml/Tests/ExchangeTrade.daml
```diff
@@ -227,4 +242,107 @@ partialExecutionsTest = do
...
+  -- Asset the DvP exchange was successful
+  [aliceDepositCid] <- queryFilter @AssetDeposit alice.customer \a -> a.asset == shareAsset
+  [bobsDepositCid] <- queryFilter @AssetDeposit bob.customer \a -> a.asset == cashAsset with quantity = 20000.0
+  -- Assert no other deposits exist for Bob
+  [] <- queryFilter @AssetDeposit bob.customer \a -> a.asset /= cashAsset with quantity = 20000.0
+  -- Assert Alice got her deposit returned to her main account
+  [(_, aliceDeposit1), (_, aliceDeposit2)] <- queryFilter @AssetDeposit alice.customer \a -> a.asset.id.label == cashAsset.id.label && a.account == alice.mainAccount
+  aliceDeposit1.asset.quantity + aliceDeposit2.asset.quantity === 980000.0
+
+  pure ()
```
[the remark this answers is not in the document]
brian-weir: Fixed

ui/src/pages/clearing/Member.tsx
```diff
@@ -97,6 +116,32 @@ const ClearingMemberComponent: React.FC<RouteComponentProps & ServicePageProps<S
       await ledger.exercise(choice, cid, {});
     };
 
+    if (!service) return <></>;
```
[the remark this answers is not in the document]
brian-weir: Done.

I agree we should review this throughout the project.

triggers/daml/AutoApproval.daml
```diff
@@ -182,19 +180,15 @@ handleApprovalRule party = do
     generateRoleCommands RegulatorRole.ApproveRegulatorRequest regulatorServicesRequests regulatorRoles ++
     generateServiceCommands RegulatorService.VerifyIdentity idVerificationRequests regulatorServices
 
-  creditRequests          <- queryByParty @Custody.CreditAccountRequest         (.provider)
-  openRequests            <- queryByParty @Custody.OpenAccountRequest           (.provider)
-  openAllocationRequests  <- queryByParty @Custody.OpenAllocationAccountRequest (.provider)
-  transferDepositRequests <- queryByParty @Custody.TransferDepositRequest       (.provider)
-  custodyServices         <- queryByParty @CustodyService.Service               (.provider)
+  depositRequests         <- queryByParty @Custody.DepositRequest   (.provider)
+  withdrawalRequests       <- queryByParty @Custody.WithdrawalRequest (.provider)
```
alexgraham-da, line 184: Slight alignment issue

triggers/test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,70 @@
+Manifest-Version: 1.0
```
alexgraham-da, line 1: What does this file do? I've never seen these before

daml/Marketplace/Clearing/Service.daml
```diff
@@ -23,27 +20,37 @@ data CalculationResult a b
   = CalculationSuccess { successCid : ContractId a, deposits: Optional DepositWithRemaining}
   | CalculationFailure { failureCid : (ContractId b)}
 
+data DepositAction
```
alexgraham-da, line 23: Is this used anywhere?

ui/daml/UI/Onboarding.daml
```diff
@@ -53,45 +47,37 @@ handleOnboardingInstruction operator customer mp ins = case ins of
...
-        void $ exercise offerCid Clearinghouse.Accept with ..
+        void $ exercise offerCid Clearinghouse.Accept with ccpAccount = account
 
     o@OnboardCustody{..} -> do
       offerCid <- exerciseByKey @Custodian.Role (operator, provider) Custodian.OfferCustodyService with ..
       void $ exercise offerCid Custody.Accept
 
     ot@OnboardTrading{..} -> do
+        let account = Custody.createAccount custodian customer
         custodyCid <- createOrLookup Custody.Service with provider = custodian; ..
```
alexgraham-da, line 62: Is creating the `Custody.Service` here still technically necessary? From what I understand with this PR is that while you will need a custody service to trade, there's not any one custody service tied to the trading service. So I can see it make sense either way - require a custodian in the instruction and create it here to ensure the customer has at least one custody service, or skip creating the custodian to avoid the confusion of the custody service not being tied to the trading service.

ui/daml/UI/Onboarding.daml
```diff
@@ -100,137 +86,66 @@ handleOnboardingInstruction operator customer mp ins = case ins of
...
+roleDependencies (OnboardIssuance p c)       = [(p, OnboardCustodian),    (c, OnboardCustodian)]
+roleDependencies (OnboardTrading e c)        = [(e, OnboardExchange),     (c, OnboardCustodian)]
+roleDependencies (OnboardClearing ch c )     = [(c, OnboardCustodian),    (ch, OnboardClearinghouse c)]
+roleDependencies (OnboardAuction ah c)       = [(ah, OnboardDistributor), (c, OnboardCustodian)]
+roleDependencies (OnboardBidding ah c)       = [(ah, OnboardDistributor), (c, OnboardCustodian)]
+roleDependencies (OnboardMarketClearing p c) = [(p, OnboardClearinghouse c )]
 roleDependencies _                           = []
 
 -- |List of observers necessary to be on accounts created for this 'OnboardingInstruction'
 accountObservers : OnboardingInstruction -> [Party]
```
alexgraham-da, line 131: This function shouldn't be necessary anymore

ui/src/pages/QuickSetup/AccountSelection.tsx
```diff
@@ -119,128 +74,29 @@ const ProviderOption = (props: {
...
-  };
   return (
     <>
       <Form.Select
         className="request-select"
         label={<p className="input-label">{accountInfo.accountLabel}</p>}
         placeholder="Select..."
         required
-        options={[...accountOptions, ...providerOptions]}
+        options={[...accountNames]}
```
Sarahb-da, line 96: `options={accountNames}` should work

ui/src/pages/QuickSetup/ProvideServicesPage.tsx
```diff
@@ -133,29 +123,20 @@ const RequestServicesPage = (props: { adminCredentials: Credentials }) => {
...
-  const allAllocationAccounts = useStreamQueries(AllocationAccountRule);
-  const allocAccounts = useMemo(
-    () => allAllocationAccounts.contracts.filter(c => c.payload.account.owner === party),
-    [allAllocationAccounts, party]
-  );
-  useEffect(
-    () => setAccountsForParty({ accounts, allocAccounts }),
-    [accounts, allocAccounts, setAccountsForParty]
-  );
+  useEffect(() => setAccountsForParty({ accounts }), [accounts, setAccountsForParty]);
```
Sarahb-da, line 139: this looks like it could cause infinite recursion..

ui/src/pages/QuickSetup/ProvideServicesPage.tsx
```diff
@@ -293,22 +222,10 @@ const RequestForm = (props: {
   const hasAccountsForServices = requestInfo?.services?.reduce((acc, svc) => {
     if (existingServices.includes(svc)) return acc;
     const accounts = requestInfo?.accounts;
+
     switch (svc) {
       case ServiceKind.CLEARING:
```
Sarahb-da, line 227: with 1 case, you probably don't need the switch anymore here

ui/src/pages/custody/Account.tsx
```diff
@@ -83,111 +54,55 @@ const Account: React.FC<ServicePageProps<Service> & AccountProps> = ({
...
+                {!isEmptySet(c.payload.lockers) && (
+                  <Popup
+                    trigger={<Icon name="lock" />}
+                    content={
+                      'Locked by ' +
+                      damlSetValues(c.payload.lockers)
+                        .map(a => getName(a))
+                        .sort()
+                        .join(', ')
+                    }
```
Sarahb-da, line 98: `<ToolTip/>` could also work here!

ui/src/pages/custody/Account.tsx
```diff
@@ -83,111 +54,55 @@ const Account: React.FC<ServicePageProps<Service> & AccountProps> = ({
...
+                  />
+                )}
               </p>
-              {party === targetAccount.account.owner && normalAccount && (
+              {party === targetAccount.account.owner && isEmptySet(c.payload.lockers) && (
                 <OverflowMenu>
                   <OverflowMenuEntry label={'Withdraw'} onClick={() => requestWithdrawDeposit(c)} />
-                  <OverflowMenuEntry label={'Transfer'} onClick={() => requestTransfer(c)} />
+                  {/*<OverflowMenuEntry label={'Transfer'} onClick={() => requestTransfer(c)} />*/}
```
Sarahb-da, line 105: comment?

ui/src/pages/custody/Assets.tsx
```diff
@@ -12,62 +10,37 @@ import { ServicePageProps, createDropdownProp } from '../common';
...
-  const onRequestCredit = async () => {
+  const onRequestDeposit = async () => {
     const asset = assets.find(i => i.payload.description === creditAsset);
-    const targetAccount = allAccounts.find(a => a.contractId === selectedAccount);
-    if (!asset || !targetAccount) return;
-    const service = clientServices.find(
-      s => s.payload.provider === targetAccount?.account.provider
-    );
 
+    if (!asset) return;
```
Sarahb-da, line 34: does this return warrant an error message?

triggers/daml/AutoApproval.daml
brian-weir: Fixed on the latest commit

daml/Marketplace/Custody/Service.daml
```diff
@@ -77,62 +43,22 @@ template Service
...
 
-      nonconsuming CreditAccount : ContractId AssetDeposit
-        with
-          creditAccountRequestCid : ContractId Custody.CreditAccountRequest
-        do
-          Custody.CreditAccountRequest{accountId; asset} <- fetchAndArchive creditAccountRequestCid
-          depositCid <- exerciseByKey @AssetSettlementRule accountId AssetSettlement_Credit with ctrl = provider; ..
-          deposit <- fetch depositCid
-          exercise depositCid AssetDeposit.AssetDeposit_SetObservers with newObservers = deposit.asset.id.signatories `union` deposit.observers
+          create AssetDeposit with asset; account; lockers = empty; observers = fromList [ registrar ]
```
GeorgSchneider, line 55: Shouldn't the observer here be the `issuer` of the asset, not the `registrar`? The registrar should already be the same as the custody provider on this service (and we should probably enforce this), and it would be a signatory on that asset deposit (and therefore see it).

daml/Marketplace/Custody/Service.daml
```diff
@@ -77,62 +43,22 @@ template Service
...
-          exerciseByKey @AssetSettlementRule accountId debit
-
-      nonconsuming TransferDeposit : ContractId AssetDeposit
-        with
-          transferDepositRequestCid : ContractId Custody.TransferDepositRequest
-        do
-          Custody.TransferDepositRequest{accountId; transfer} <- fetchAndArchive transferDepositRequestCid
-          exerciseByKey @AssetSettlementRule accountId transfer
+          Custody.WithdrawalRequest{depositCid} <- fetchAndArchive withdrawalRequestCid
+          archive depositCid
```
GeorgSchneider, line 62: I think we need to assert here as well that the account provider on the asset deposit is in fact the provider of the custody service we're in. The archive would fail if it's not, but with the assert we can provide a proper error message.

daml/Marketplace/Custody/Service.daml
```diff
@@ -190,4 +117,11 @@ template Request
         operator : Party
       controller operator, provider
       do
+        let account = createAccount provider customer
         createOrLookup Service with ..
+
+createAccount : Party -> Party -> Account
+createAccount provider customer = do
+  let id = Id with signatories = fromList [customer, provider]; label = partyToText customer <> "@" <> partyToText provider; version = 0
+      owner = customer
```
GeorgSchneider, line 126: Comment for other reviewers: this is - for now - the default account naming we agreed upon. We can consider re-adding user-defined account naming (or in fact multiple account support) later again, if we see the need for it.

daml/Marketplace/Distribution/Auction/Service.daml
```diff
@@ -91,7 +86,7 @@ template Service
           (remainingDepositCidOpt, siCids) <- foldlA generateSettlementInstruction (Some depositCid, []) settleAllocations
 
           -- Transfer remaining deposit back to the Issuer's trading account
-          whenSome remainingDepositCidOpt \depositCid -> (exerciseByKey @AllocationAccountRule allocationAccount.id Allocation.Withdraw with depositCid = depositCid; transferTo = tradingAccount) $> ()
```
GeorgSchneider, line 94: Improve the comment here: we're not transferring the remaining deposit, but just unlocking it again

daml/Marketplace/Issuance/Service.daml
```diff
@@ -57,13 +55,14 @@ template Service
...
           -- assertMsg "issued must be 0" $ description.issued == 0.0
           -- assertMsg "authorized must be > 0" $ description.authorized > 0.0
           archive createOriginationCid
-          let
-            assetId = Id with signatories = fromList [ provider, customer ]; label = assetLabel; version = 0
-            issuer = customer
+          let assetId = Id with signatories = fromList [ provider, customer ]; label = assetLabel; version = 0
+              issuer = customer
+              registrar = provider
+
```
GeorgSchneider, line 65: Note: implicit assumption here that the provider of the issuance service is the same as the registrar (root custodian) of the asset at hand. I think it's reasonable to assume that but we should think about possible situations where this might not be the case (and if / how it can be supported)

daml/Marketplace/Issuance/Service.daml
```diff
@@ -85,27 +84,22 @@ template Service
...
+          CreateIssuanceRequest{issuanceId; assetId; account; quantity} <- fetchAndArchive createIssuanceRequestCid
           issuanceCid <- create Issuance with ..
 
+          (_, AssetDescription{registrar}) <- fetchByKey @AssetDescription assetId
           let asset = Asset with id = assetId; quantity
-          depositCid <- exerciseByKey @AssetSettlementRule accountId AssetSettlement_Credit with ctrl = provider; ..
-          deposit    <- fetch depositCid
-          depositCid <- exercise depositCid AssetDeposit.AssetDeposit_SetObservers with newObservers = deposit.asset.id.signatories `union` deposit.observers
+          depositCid <- create AssetDeposit with account; asset; lockers = empty; observers = fromList [ registrar ]
+
```
GeorgSchneider, line 93: Same comment as in `CustodyService` - should the observer be the `issuer` here?

daml/Marketplace/Settlement/Model.daml
```diff
@@ -29,19 +26,5 @@ template SettlementInstruction
...
-
-            let
-              isCtrl         = member provider rule.ctrls
-              addProvider    = unless isCtrl . void $ exerciseByKey @AssetSettlementRule transferTo.id AssetSettlement_AddController with ctrl = provider
-              removeProvider = unless isCtrl . void $ exerciseByKey @AssetSettlementRule transferTo.id AssetSettlement_RemoveController with ctrl = provider
-              settle         = exerciseByKey @AllocationAccountRule settlementDetails.senderAccount.id Transfer with ..
-
-            -- Sequence the transaction, ignoring the output of the add and remove provider transactions but return the output of the settle transaction
-            addProvider *> settle <* removeProvider
-            )
```
GeorgSchneider, line 47: I love how that part is greatly simplified!

daml/Marketplace/Trading/Service.daml
```diff
@@ -1,99 +1,85 @@
...
+      unlockCollateral : Order.MarketType -> Update ()
+      unlockCollateral = \case
+        Order.Collateralized depositCid _ -> unlockDeposit depositCid
+        _ -> pure ()
 
     controller customer can
-      nonconsuming RequestCreateOrder : Either (ContractId Order.T)
-                                               (ContractId Order.T, ContractId CreateOrderRequest, Order.TradeCollateral)
+      nonconsuming RequestCreateOrder : ContractId Order.T
         with
```
GeorgSchneider, line 38: Any reason why we don't (need to?) return the `ContractId CreateOrderRequest` here anymore? I'm thinking it might come in handy in test scripts, where we need to pass it into the next choice.

daml/Marketplace/Trading/Service.daml
```diff
@@ -1,99 +1,85 @@
...
-                None -> do
-                  let reason = Error with code = "3001"; message = "Customer is not a member of clearinghouse"
-                  Left <$> create Order.Order with status = Order.Rejected{reason}; ..
-
-                (Some serviceCid) -> exercise serviceCid Clearing.ApproveTrade >>= \case
+            detailsWithDepositedFees : Update Order.Details
+            detailsWithDepositedFees = case details.optExchangeFee of
+              Some feeDepositCid -> do
+                newDepositCid <- exercise feeDepositCid AssetDeposit_Lock with newLockers = Set.fromList [ provider ]
+                return details with optExchangeFee = Some newDepositCid
```
GeorgSchneider, line 54: Do I understand this correctly here: we lock the entire fee amount that the user has provided, and then unlock and transfer parts of it with every execution?

I don't see anywhere that we check the provided fee amount is actually sufficient to execute the full order. Don't need to add this in this PR, but maybe open a ticket for it.

daml/Marketplace/Trading/Service.daml
GeorgSchneider: I just saw further down that in the error case we just create an `Order` contract, which is probably the reason for this change in choice signature. I wonder if we should just `fail` in case something is wrong with the order on entry.

GeorgSchneider: (but ^ is also not something to do in this PR)

daml/Marketplace/Trading/Service.daml
```diff
@@ -123,24 +109,23 @@ template Service
...
-          orderCid <- create order with status = Order.Rejected with reason
-          collateral <- withdrawFromAllocationAccount collateral
-          optExchangeFee <- withdrawFeesFromAllocationAccount details.optExchangeFee
 
-          return (orderCid, collateral)
+          create order with status = Order.Rejected with reason
 
-      nonconsuming MarketOrderCancelRequest : (ContractId Order.T, Order.TradeCollateral)
+      nonconsuming MarketOrderCancelRequest : ContractId Order.T
         with
```
GeorgSchneider, line 129: Not related to this PR, but: why do we need this choice, besides the `CancelOrder` one?

daml/Marketplace/Trading/Service.daml
```diff
@@ -285,14 +250,14 @@ template CreateOrderRequest
         do
           exerciseByKey @Service serviceKey AcknowledgeOrderRequest with createOrderRequestCid = self, ..
 
-      nonconsuming RejectRequest : (ContractId Order.T, Order.TradeCollateral)
+      nonconsuming RejectRequest : ContractId Order.T
         with
```
GeorgSchneider, line 254: Again, question unrelated to this PR: What's the reason we need this and the below choices on this template (as a mere proxy to the choices on the service)?

daml/Tests/Bond.daml
```diff
@@ -1,34 +1,29 @@
 module Tests.Bond where
 -- ^ This module tests the full lifecycle of a binary option
```
GeorgSchneider, line 3: `bond` instead of `binary option`

triggers/test/META-INF/MANIFEST.MF
GeorgSchneider: Some build artifact? Prob can be removed

ui/daml/UI/Onboarding.daml
GeorgSchneider: Agree, I think this was needed to get the necessary accounts for the trading service, but that's not required anymore. So I *think" it should be safe to remove, but check if there's some implicit dependency on this being created here

ui/daml/UI/Onboarding.daml
```diff
@@ -100,137 +86,66 @@ handleOnboardingInstruction operator customer mp ins = case ins of
...
-        receivableAccount <- createOrLookupAccount    ot custodyCid optReceivableAccount "Auction-ReceivableAccount"
-        allocationAccount <- createAllocationAccount' ot custodyCid "allocation"
-
         offerCid <- exerciseByKey @Distributor.Role (operator, provider) Distributor.OfferAuctionService with ..
-        void $ exercise offerCid Auction.Accept with ..
+        void $ exercise offerCid Auction.Accept
 
     ot@OnboardBidding{..} -> do
+        let account = Custody.createAccount custodian customer
         custodyCid <- createOrLookup Custody.Service with provider = custodian; ..
```
GeorgSchneider, line 98: Same question as above: is creating a custody service needed here, now that we have removed the account dependency?

GeorgSchneider (approved):
I've left a couple of comments / questions, but looks good to me overall 👍

triggers/test/META-INF/MANIFEST.MF
brian-weir: I removed it - it was generated by the daml compiler but I'm not sure why...

daml/Marketplace/Clearing/Service.daml
brian-weir: Well spotted, I used it in a previous incarnation but forgot to remove it. 

It's removed in the latest commit.

daml/Marketplace/Trading/Service.daml
alexgraham-da: This is currently checked in the Exchange Adapter / Matching Engine. The reason for that as a workaround was without `submitMulti`, the `customer` exercising this choice has no way of being able to look up the `FeeSchedule` to get the fee amount.

The check can be moved here when we can start using `submitMulti` with Public party reads. I know that @alexmatson-da has been working on supporting this in hub - is it integrated into the frontend yet?

ui/src/pages/QuickSetup/AssignRolesPage.tsx
```diff
@@ -366,45 +313,36 @@ const makeInstruction = (inst: InstFieldsWithType): OnboardingInstruction => {
 const getFields = (inst: InstFieldsWithType) => {
   const provider = FieldType.PARTIES;
   const custodian = FieldType.PARTIES;
-  const tradingAccount = FieldType.TEXT;
   const clearingAccount = FieldType.TEXT;
```
Sarahb-da, line 316: can `clearingAccount` be removed too?

ui/src/pages/QuickSetup/ProvideServicesPage.tsx
```diff
@@ -458,57 +375,24 @@ const CreateServiceRequests = (props: {
...
-              if (!clearingAccounts?.clearingAccount || !clearingAccounts?.marginAccount) {
+              if (!clearingAccounts?.clearingAccount) {
                 return;
               }
               const clearingParams = {
                 ...params,
                 clearingAccount: clearingAccounts.clearingAccount,
-                marginAccount: clearingAccounts.marginAccount,
               };
               await ledger.create(ClearingRequest, clearingParams);
```
Sarahb-da, line 385: this can be a `doRequest` right?

Sarahb-da: Only would prefer that if a `displayErrorMessage` could also be added to that function, uniformly handling failed requests for all

ui/src/pages/QuickSetup/QuickSetup.tsx
```diff
@@ -131,12 +130,6 @@ const QuickSetup = withRouter((props: RouteComponentProps<{}>) => {
                     <ArrowRightIcon />
                   </Button>
                 </NavLink>
-                <NavLink to={`${matchUrl}/${MenuItems.CREATE_ACCOUNTS}`}>
```
Sarahb-da, line 134: line 35 in this file can also probably go

ui/src/pages/clearing/AllocationModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { CreateEvent } from '@daml/ledger';
+import {
+  AssetDeposit,
+  AssetDeposit_SetObservers,
+  AssetDeposit_Split,
+} from '@daml.js/da-marketplace/lib/DA/Finance/Asset';
+import { Button, Form, Header, Modal } from 'semantic-ui-react';
+
+type MarginCallProps = {
+  deposit: CreateEvent<AssetDeposit> | undefined;
```
Sarahb-da, line 12: or `deposit?: CreateEvent<AssetDeposit>`

ui/src/pages/clearing/AllocationModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+  observers,
+}) => {
+  const ledger = useLedger();
+  const [allocation, setAllocation] = useState<number>(0.0);
+
+  useEffect(() => {
+    if (!!deposit) setAllocation(parseFloat(deposit.payload.asset.quantity));
+  }, [open, deposit]);
+
+  if (!deposit) return <></>;
```
Sarahb-da, line 33: better to `return null` so that you don't render anything from this component in this case

ui/src/pages/clearing/AllocationModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const requestAllocation = async () => {
+    const depositQuantity = parseFloat(deposit.payload.asset.quantity);
+
+    if (allocation <= 0 || allocation > depositQuantity) {
+      onClose(false);
+      return;
```
Sarahb-da, line 40: error message here? "The allocated quantity must be greater than or equal to 0 and less than the deposit"

ui/src/pages/clearing/Member.tsx
```diff
@@ -33,22 +39,35 @@ const ClearingMemberComponent: React.FC<RouteComponentProps & ServicePageProps<S
...
-
     const standing = standings.find(standing => standing.payload.customer === customer);
+
+    const { contracts: allDeposits, loading: depositsLoading } = useStreamQueries(AssetDeposit);
+    const deposits = allDeposits.filter(
+      a => a.payload.account.id.label === service?.clearingAccount.id.label
+    );
+    const availableDeposits = deposits.filter(
+      d => isEmptySet(d.payload.lockers) && !d.payload.observers.map.has(service?.provider || '')
+    );
```
Sarahb-da, line 65: `map.has`.. nice!

daml/Marketplace/Trading/Service.daml
alexmatson-da: Yep being able to retrieve a token with `readAs: public-id` is now supported in Hub, which should allow the submitMulti. It's not yet integrated into the marketplace frontend code, though

alexgraham-da (approved):
This is great! Really streamlines things and simplifies the codebase.

ui/src/pages/distribution/bidding/Auction.tsx
```diff
@@ -74,8 +88,7 @@ export const BiddingAuction: React.FC<ServicePageProps<Service>> = ({
     render(el2.current, data);
   }, [el2, assets, biddingAuction, showQuotedAsset]);
 
-  if (!biddingAuction || services.length === 0) return <></>;
-  const service = services.filter(s => s.payload.customer === party)[0];
+  if (!biddingAuction || !service) return <></>;
```
Sarahb-da, line 91: another place u can just return `null`

ui/src/pages/notifications/NotificationComponents.tsx
```diff
@@ -152,7 +152,8 @@ export function RequestNotification<T extends Fields<{ operator: string }>>({
   const onApprove = async () => {
     if (lookupFields) {
       const args = lookupFields(approveArgs);
-      ledger.exercise(approveChoice, contract, args);
+      console.log('args:' + args.operator);
```
Sarahb-da, line 155: extra log?

ui/src/pages/origination/NewBaseInstrument.tsx
```diff
@@ -60,25 +70,19 @@ const NewBaseInstrumentComponent = ({ history }: RouteComponentProps) => {
...
-    const safekeepingAccount = accounts.find(
-      a => a.provider === service.payload.provider && a.id.label === account
-    );
-    if (!safekeepingAccount) {
-      console.log(
-        `Couldn't find account from provider ${service.payload.provider} with label ${account}`
-      );
+    const service = customerServices.find(i => i.payload.provider === registrar);
+    if (!service) {
+      console.log(`Couldn't find issuance service for selected registrar ${registrar}`);
```
Sarahb-da, line 78: extra log?

ui/src/pages/origination/NewConvertibleNote.tsx
```diff
@@ -117,27 +122,22 @@ const NewConvertibleNoteComponent = ({ history }: RouteComponentProps) => {
...
-    const safekeepingAccount = accounts.find(
-      a => a.provider === service.payload.provider && a.id.label === account
-    );
-    if (!safekeepingAccount) {
-      console.log(
-        `Couldn't find account from provider ${service.payload.provider} with label ${account}`
-      );
+    const service = customerServices.find(i => i.payload.provider === registrar);
+    if (!service) {
+      console.log(`Couldn't find issuance service for selected registrar ${registrar}`);
```
Sarahb-da, line 130: extra log?

Sarahb-da (approved):
a few minor comments. otherwise looks good!

ui/daml/UI/Onboarding.daml
brian-weir: I actually had the same thought but I left it in. It does look strange in the quick setup as for some roles you have to specify the `Custody` service provider multiple times. 

I do think we should review this but perhaps outside of this PR, if you agree ?

ui/src/pages/QuickSetup/AccountSelection.tsx
brian-weir: Updated, thanks!!

ui/src/pages/QuickSetup/ProvideServicesPage.tsx
brian-weir: Fixed

ui/src/pages/custody/Account.tsx
brian-weir: Removed

ui/src/pages/custody/Account.tsx
brian-weir: Not sure what you're referring to here with `<ToolTip/>` as I cannot find it in the project. I only see `<div className="tool-tip">Click here to fit to screen</div>` which is on the ReviewPage. Are you referring to this ?

ui/src/pages/custody/Assets.tsx
brian-weir: We could refactor the code to avoid having to do a `find` on the list of available assets so we don't need this check. I'd prefer to do that as part of a separate PR if that's ok?

daml/Marketplace/Custody/Service.daml
brian-weir: I've added the assertion in the latest commit.

ui/src/pages/custody/Account.tsx
Sarahb-da: Nope. My mistake - when I suggested this I searched for the term in the project and got a false positive thinking that _that_ instance of ToolTip was a common one - never mind!

daml/Marketplace/Issuance/Service.daml
brian-weir: This is a very important point - because we also make this assumption in multiple places in the UI when determining the correct `Custody.Service` to utilise or when determining the correct delivery account. We assume the `Issuance.Service` provider is the `registrar` on the `AssetDescription` and inherently the `Custody.Service` provider are the same party.

daml/Marketplace/Issuance/Service.daml
brian-weir: Updated in the latest commit - Yes, this is so the `issuer` has visibility on parties which holds assets the `issuer` has created.

daml/Marketplace/Trading/Service.daml
brian-weir: > Any reason why we don't (need to?) return the `ContractId CreateOrderRequest` here anymore? I'm thinking it might come in handy in test scripts, where we need to pass it into the next choice.

I presume it was so we have easy access to it when testing but it was actually never referenced in any testing scripts nor in the entire project which was why I removed it. 

> I just saw further down that in the error case we just create an `Order` contract, which is probably the reason for this change in choice signature. I wonder if we should just `fail` in case something is wrong with the order on entry.

Great point - handling failure/error scenarios is something we need to review in the entire Marketplace. I believe the error cases were added so it would be easy for the investor to see why an order has not been submitted for execution as there was no other way of "nicely" displaying an error message back to the user at that point in time. 

> (but ^ is also not something to do in this PR)

👍

daml/Marketplace/Trading/Service.daml
brian-weir: This is specific to how Exberry handles Market orders - when a Market order is submitted as `Fill Or Kill` or `Immediate Or Cancelled`, when Exberry can neither fill the order nor provide a partial fill then it submits a `Order Cancel Message` on its Market Data API (see "Order Cancel Message" at https://docs.exberry.io/market-data#response).

The `MarketOrderCancelRequest` choice can only be called via the original `Trading.CreateOrderRequest` template from the Exberry Adapter (the `CreateOrderRequest` contract is created to trigger the Exberry Adapter and gets consumed based on the response from Exberry). 

The `CancelOrder` choice is called when the user requests to cancel an order which kicks off an entirely different flow to and from Exberry (for example, in Exberry user initiated cancellations are send and received on a different endpoint from the Market Orders which get "cancelled").

daml/Tests/Bond.daml
brian-weir: updated

ui/daml/UI/Onboarding.daml
brian-weir: I don't think its needed anymore as it adds confusion (especially in the UI). I'll remove it but it does require a bit of effort to do so on the UI side.

ui/src/pages/QuickSetup/AssignRolesPage.tsx
brian-weir: `ClearingAccount` is the only remaining account at the `Service` level. Luciano is working on clearing which was why I tried to make the least changes to clearing as possible

ui/src/pages/QuickSetup/ProvideServicesPage.tsx
brian-weir: > this can be a `doRequest` right?

`ClearingAccount` is required for the clearing service which is why `doRequest` won't work in its current form.

> Only would prefer that if a `displayErrorMessage` could also be added to that function, uniformly handling failed requests for all

I agree but I think this should be part of a separate PR.

ui/src/pages/clearing/AllocationModal.tsx
brian-weir: done

ui/src/pages/clearing/Member.tsx
brian-weir: The daml `Set` which is a daml `Map` with only keys exposes this `has` function :)

ui/src/pages/clearing/AllocationModal.tsx
brian-weir: I could try and convert this over to `ModalFormErrorHandled` so we display the error message to the user? Would you want this as part of this PR or another one ?

ui/src/pages/origination/NewBaseInstrument.tsx
brian-weir: removed

ui/src/pages/origination/NewConvertibleNote.tsx
brian-weir: Removed - we originally had a log here where variables where `undefined` but I agree that we don't need it (see line 129)

PR #422, opened 2021-09-02
Test clearing trigger

alexgraham-da:
add a clearing trigger test, add automation readme, add automation tests to version tagging

PR #424, opened 2021-09-02
add onboarding instruction daml to ui folder

alexgraham-da:
Adds a new daml model in the `ui` folder that is designed to make onboarding via the quick setup easier and more streamlined.

By using a list of `OnboardingInstruction`s which designate which role or service you want for a party, the `OperatorOnboarding` contract will collect the necessary signatures from the providers (via a `PartyOnboarding` contracts each party should have), create necessary roles for the providers, create needed accounts with the correct observers, and create the services. These lists can be reused for different parties to create the same set of roles/services for each party.


Sarahb-da (commented):
did u intentionally not include the git ignore change? I believe we added `daml-ui.js` and `ui/.daml`

Sarahb-da (approved):

alexgraham-da:
> did u intentionally not include the git ignore change? I believe we added `daml-ui.js` and `ui/.daml`

nope, I will add it back

alexgraham-da:
Note, to avoid merge conflicts with the `tag-versions.sh` file in https://github.com/digital-asset/da-marketplace/pull/422, I needed to pull in those changes into this branch, so that PR should be merged first.

PR #425, opened 2021-09-02
v2: update for hub react

alexmatson-da:


ui/package.json
```diff
@@ -1,91 +1,91 @@
...
-    "sass": "^1.32.8",
-    "typescript": "^3.9.5"
-  }
+    "name": "da-marketplace-ui",
+    "version": "0.2.0-rc.3",
+    "private": true,
+    "dependencies": {
+        "@daml.js/da-marketplace": "file:../daml.js/da-marketplace-0.2.0",
+        "@daml-ui.js/da-marketplace-ui": "file:../daml-ui.js/da-marketplace-ui-0.2.0",
+        "@daml/hub-react": "file:../../dabl-react",
```
alexmatson-da, line 8: I tested this whole PR with hub-react built off of this commit https://github.com/digital-asset/dabl-react/pull/31 

After that PR is merged upstream, I'll cut a `1.0.0-rc.4` release of hub-react and install it here

ui/src/pages/QuickSetup/AddPartiesPage.tsx
```diff
@@ -170,45 +181,60 @@ const AddPartiesPage = () => {
...
+  const completedPartyOnboarding = usePartyOnboarding(party, operator);
+
+  useEffect(() => {
+    if (completedIdentityVerifications && completedPartyOnboarding) {
+      return onComplete();
+    }
+  }, [completedIdentityVerifications, completedPartyOnboarding, onComplete]);
+
+  return null;
+};
```
alexmatson-da, line 199: Before there was one big useEffect doing multiple things - I split up the jobs into smaller hooks for easier management. Each hook eventually reports the completion status of that job

ui/src/util.ts
```diff
@@ -21,3 +21,24 @@ export function preciseInputSteps(precision: number): PreciseInputSteps {
...
+
+export function cache(options?: { permanent: boolean }) {
+  const store = options?.permanent ? localStorage : sessionStorage;
+
+  return {
+    save: (key: string, value: string) => store.setItem(key, value),
+    remove: (key: string) => store.removeItem(key),
+    load: (key: string) => store.getItem(key),
+  };
+}
```
alexmatson-da, line 44: Added this cache utility that can be configured with either localStorage or sessionStorage, and maintains the same backing store for each save, remove, and load operation

ui/src/pages/common.ts
```diff
@@ -54,3 +58,63 @@ export function damlSetValues<T>(damlSet: DamlSet<T>): T[] {
...
+    } else {
+      if (!publicParty && hubPublicParty) {
+        save(PUBLIC_PARTY_ID_KEY, hubPublicParty);
+        setPublicParty(hubPublicParty);
+      }
+    }
+  }, [publicParty, hubPublicParty, save]);
+
+  return publicParty;
+}
```
alexmatson-da, line 90: More abstractions to smooth over differences in Hub and local envs!

ui/src/Credentials.ts
```diff
@@ -58,19 +60,23 @@ export function retrieveCredentials(): Credentials | undefined {
 
 export function computeToken(party: string): string {
   const payload = {
```
Sarahb-da, line 62: since this is used locally only, may be worth specifying that in the name

ui/src/pages/QuickSetup/AccountSelection.tsx
```diff
@@ -144,8 +144,27 @@ const ProviderOption = (props: {
...
+  const serviceProviderName = getName(serviceProvider);
+
+  const makeAccountName = useCallback(
+    (accountInfo: AccountInfo): string | undefined => {
+      const name = `${getName(party)}-${getName(
+        serviceProvider
+      )}-${accountInfo.accountLabel.replace(/\s+/g, '')}`;
+
+      if (isHubDeployment) {
+        if (partyName !== party && serviceProviderName !== serviceProvider) {
```
Sarahb-da, line 157: if u already know it is a hub deployment, do u still need to check this?

alexmatson-da: yeah - the difference between hub and local is the long party IDs. on local, the party ID is usually human readable anyway 

On Hub, I want to enforce that the `partyName` is different from the actual party ID before using it in the account name label (this could happen, because using `getName(party)` will default to party ID if no other suitable name is found, like if the verifiedIdentities contracts haven't loaded in yet)

ui/src/pages/clearing/ClearingOfferModal.tsx
```diff
@@ -85,7 +85,7 @@ const ClearingOfferModal: React.FC<ServicePageProps<Service> & OfferProps> = ({
     const accountRequest: RequestOpenAccount = {
       accountId: {
         signatories: makeDamlSet([service.payload.provider, service.payload.customer]),
-        label: `${party}-${clearingProvider}-clearing`,
+        label: `${getName(party)}-${getName(clearingProvider)}-clearing`,
         version: '0',
```
Sarahb-da, line 89: This reminds me of `makeAccountName` where u check `isHubDeployment`, but you don't do it here.. curious why?

ui/src/pages/common.ts
```diff
@@ -54,3 +58,63 @@ export function damlSetValues<T>(damlSet: DamlSet<T>): T[] {
...
+
+export function usePublicParty() {
+  // Cache in localStorage to share across all tabs & restarts
+  const { save, load } = React.useMemo(() => cache({ permanent: true }), []);
+
+  const [publicParty, setPublicParty] = React.useState<string>();
+  const hubPublicParty = useHubPublicParty();
+
+  React.useEffect(() => {
+    const cachedPublicParty = load(PUBLIC_PARTY_ID_KEY);
```
Sarahb-da, line 72: do u need to check if `!load` here?

ui/src/pages/login/Login.tsx
```diff
@@ -234,10 +201,12 @@ const PartiesLoginForm: React.FC<Props> = ({ onLogin }) => {
     }
   };
 
-  const handleLoad = async (parties: PartyDetails[]) => {
-    setParties(parties);
-    setSelectedPartyId(parties[0]?.party || '');
-    storeParties(parties);
+  const handleLoad = async (parties: PartyToken[]) => {
```
Sarahb-da, line 204: If you want, you could re-use some of this on line 188. Looks like you check and set parties + selectedPartyId in the same way

ui/src/pages/login/Login.tsx
```diff
@@ -250,24 +219,38 @@ const PartiesLoginForm: React.FC<Props> = ({ onLogin }) => {
...
-                  <Icon name="file" className="white" />
-                  <p className="dark">Load Parties</p>
-                </label>
+                <DamlHubLogin
+                  options={{
+                    method: {
+                      file: {
+                        render: () => (
+                          <label className="custom-file-upload button ui">
+                            {' '}
```
Sarahb-da, line 237: these sneak in sometimes...

alexmatson-da: I've noticed this too. It's a thing that Prettier does to preserve spaces, if you have some JSX/HTML that spans multiple lines, then whitespace is collapsed. So something like 

```
<p>Hello 
    <a>world</a>
</p>
```

would render on the browser as `Helloworld`. Thus Prettier would change that to 

```
<p>
    Hello {' '} 
    <a>world</a>
</p>
```

so the explicit space is preserved. In _this_ case, there's no need for it because there's no text at the start, but I probably had a whitespace there in the code before the formatting, and Prettier overzealously preserved it (you see it happened to this guy too https://stackoverflow.com/questions/64358015/why-does-prettier-automatically-insert-empty-interpolation-code-characters)

ui/src/pages/common.ts
alexmatson-da: it's always defined

ui/src/pages/clearing/ClearingOfferModal.tsx
alexmatson-da: yep, same thing... I'll reuse that function

Sarahb-da (approved):
lgtm, nice work

ui/src/Credentials.ts
```diff
@@ -58,19 +60,23 @@ export function retrieveCredentials(): Credentials | undefined {
...
-      ledgerId: ledgerId,
+      ledgerId: 'da-marketplace-sandbox',
       applicationId: APPLICATION_ID,
       actAs: [party],
-      readAs: [party, publicParty],
+      readAs: [party, 'Public'],
     },
+    party,
+    ledgerId: 'da-marketplace-sandbox',
+    partyName: party,
```
brian-weir, line 72: Just wondering why you're hardcoding these values ? I presume they were earlier just static variables from some config file.

brian-weir (approved):
lgtm ! Great work Alex!

ui/src/Credentials.ts
alexmatson-da: In short, simply because this function to compute a token is only ever run locally (dev), and these values will always be the same for dev. 

On top of that, we don't need to know the `ledgerId` anywhere else in the app anymore (on Hub, the JSON API is now relative to the base domain), and the `usePublicParty()` hook is used to dynamically fetch the correct public ID in whatever running environment the app is in

brian-weir: Thanks!

PR #431, opened 2021-09-20
Fix operator role streaming and dropdown value on Hub

alexmatson-da:


ui/src/components/InputDialog/Fields.tsx
```diff
@@ -12,7 +12,11 @@ interface RegularField {
 interface SelectionField {
   label: string;
   type: 'selection';
-  items: string[];
+  items: DropdownItemProps[];
+}
+
+export function defaultItems(items: string[]): DropdownItemProps[] {
```
Sarahb-da, line 18: can this be 'formatDefaultDropdownItems' or something? 'defaultItems' is pretty vague

Sarahb-da (approved):
lgtm

ui/src/components/InputDialog/Fields.tsx
alexmatson-da: made it so that both `string[]` and `DropdownItems[]` are allowed as `items`, and it internally converts the string array to dropdownItemProps, if needed 

that way an external function isn't necessary

PR #433, opened 2021-09-23
0.1 fix settlement bug

alexgraham-da:
Fix a bug where if the buyer bid for more than the trade was filled at, the entire collateral would be transferred instead of just the portion owed.


alexmatson-da (commented):
lgtm

alexmatson-da (approved):

PR #437, opened 2021-09-27
Matching engine upgrade

brian-weir:
PR includes two fixes - 

1. Returning unused `AssetDeposit`'s for over collateralised orders;
2. Updates to the matching engine to assign the correct price to limit orders for the scenario's laid out by Eric :

> scenario 1:
> Seller : Resting offer of 1 @ 40
> Buyer : Pay 50 on 1
> Result : Seller sold 1 @ 50 to Buyer (edited) 
> **Expected Result - Executed at 40 (confirmed on Exberry)**
> 
> scenario 2:
> Buyer : Resting 50b for 1
> Seller : Sell 1 at 40
> Result : Seller sold 1 @ 50 to Buyer
> **Expected result - Executed at 50b (confirmed on Exberry)**
> 
> scenario 3:
> Seller : Resting offers for 1 @ 40, 45, and 50
> Buyer : Pay 50 on 3
> Result : Seller sold 3 @ 50 to Buyer
> **Expected result -
>   3 Separate executions - 1 unit @ 40, 45 and 50
>       (confirmed on Exberry)**


triggers/daml/MatchingEngine.daml
```diff
@@ -158,20 +154,18 @@ matchOrder party orders op@(orderCid, order) = do
 
     fill : ContractPair Order.T -> ContractPair Order.T -> TriggerA MatchingState ()
     fill (aggressiveCid, aggressive) (passiveCid, passive) = do
-      debug $ "Matching order: " <> show aggressive.details.id <> " to : " <> show passive.details.id
+      debug $ "Matching order: Taker=" <> show aggressive.details.id <> " to Maker" <> show passive.details.id
```
alexgraham-da, line 157: should probably be `to Maker="`

triggers/test/daml/Tests/Matching.daml
```diff
@@ -56,13 +66,14 @@ runMatchingTest parties@Parties{userAdmin = operator; ..} = do
     observers = []
   createListingRequestCid <- alice.customer `submit` do exerciseCmd listingServiceCid Listing.RequestCreateListing with ..
 
-  debug "Waiting for listing to be accepted..."
+  -- debug "Waiting for listing to be accepted..."
```
alexgraham-da, line 69: Can either decomment or delete this one (since it already has the "Creating listing..." message earlier)

triggers/test/daml/Tests/Matching.daml
```diff
@@ -90,9 +101,170 @@ runMatchingTest parties@Parties{userAdmin = operator; ..} = do
             side = Order.Buy
             optExchangeFee = None
             timeInForce = Order.GTC
+            orderType = Order.Limit with price = 150.0
+
+  debug "Waiting for orders to be executed..."
```
alexgraham-da, line 106: Does running this test with the Exberry integration still pass with these additions? The test was initially intended to pass with both the matching engine and the integration - though had it been this detailed originally, we probably would have caught the matching engine issue fixed in this PR earlier!

alexgraham-da (approved):
Looks good to me, though I would like to know if the new test passes with the Exberry integration. If it doesn't pass, that may just mean Exberry is handling matching logic different to us, which is fine. In that case though, we would still need the original version of this test to exist as a way to test the Exberry adapter. I wrote unit tests for the integration itself in the integration's repo, but it doesn't cover the adapter's functionality.

PR #439, opened 2021-10-07
install prettier import sort plugin

alexmatson-da:


Sarahb-da (approved):
U can prob remove the current import sort from the make file right??

PR #442, opened 2021-10-08
0.1.x security

alexgraham-da:
- Require Custodian to have a relationship with issuer to deposit tokens created by Issuer
- Adds locking version of `AssetDeposit`s and removes all `Marketplace.Transfer` requests from trading. This allows trades to be fully settled and collateralized without having the custodian trigger running
- Require custodian signature on requests so that rogue parties cannot create requests without a custodian relationship
- Add Operator as signatory on `AssetDeposit` and other contracts to avoid rogue custodian setups

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,7 +212,7 @@ $(ui): $(dar) $(exberry_adapter)
 
 .PHONY: clean
 clean: clean-ui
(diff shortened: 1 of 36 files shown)


brian-weir (approved):
Nice work, a lot of changes but it looks good! 

Apologies for the late review!

PR #459, opened 2022-01-21
Syndication - next iteration

GeorgSchneider:
Apologies for the huge size - this PR covers: FL-34, FL-173, FL-42, FL-41, FL40, FL-169, FL-171.

Best reviewed by going through each commit related to a given ticket.

Commit for FL-137 and FL-138 is not tagged with the ticket number. It's this one: `Support multiple accounts per provider and selection of settlement ac...`


daml/Marketplace/Lifecycle/Model.daml
```diff
@@ -140,7 +140,7 @@ template Effect
                 let
                   settlementId = lifecycleId <> "-" <> partyToText snapshot.deposit.account.owner <> "-Expiry"
                   asset = snapshot.deposit.asset with id.signatories = fromList [bondRegistrar, issuer]; id.version = newVersion
-                instructionCids <- createInstructions False operator payingAgent bondRegistrar snapshot.deposit.account.owner issuer settlementId 0 asset
+                instructionCids <- createInstructions False operator payingAgent bondRegistrar snapshot.deposit.account.owner bondRegistrar settlementId 0 asset
```
judywu-da, line 141: <placeholder>

daml/Tests/Distribution/Syndication/Lifecycle.daml
```diff
@@ -95,10 +95,10 @@ lifecycle Parties{..} date assetId = do
       provideSnapshot custodian5 maxId date
       paymentCids <- instructLifecycle operator payingAgent maxId date
 
-      allocateInstructions [ issuer, lm1, lm2, lm3, custodian1, custodian2, custodian4, custodian5, investor1, investor2, investor3, investor4 ]
-      signInstructions [ issuer, lm1, lm2, lm3, custodian1, custodian2, custodian4, custodian5, investor1, investor2, investor3, investor4 ]
+      allocateInstructions [ issuer, bondRegistrar, lm1, lm2, lm3, custodian1, custodian2, custodian4, custodian5, investor1, investor2, investor3, investor4 ]
```
judywu-da, line 98: <remove this later>

daml/Marketplace/Custody/Service.daml
```diff
@@ -12,17 +12,25 @@ template Service
...
+        with
+          label : Text
+        do
+          assertMsg "Account label cannot be empty" $ label /= ""
+          create Custody.OpenAccountRequest with ..
+
       nonconsuming RequestDeposit : ContractId Custody.DepositRequest
         with
           asset : Asset
+          account : Account
```
judywu-da, line 33: @dorrit-du-da let's inform KC about this change as it will affect asset bridge

daml/Tests/Distribution/Syndication/Lifecycle.daml
GeorgSchneider: True, good spot

daml/Marketplace/Lifecycle/Model.daml
GeorgSchneider: Yes, that's correct. It was the pragmatic solution, as there's no need for the issuer to receive the redeemed bonds anyway.

daml/Marketplace/Custody/Service.daml
GeorgSchneider: Yes, that's correct - now there's no "default account" with a custodian anymore. I suppose one would need to select one for asset bridge deposits, and could use the `AccountInfo` contracts for that.

judywu-da:
@dorrit-du-da @georg-da I have done reviewing commits https://github.com/digital-asset/da-marketplace/pull/459/commits/35403fa46e00a087bec2a809bbf211bd55c19781 and https://github.com/digital-asset/da-marketplace/pull/459/commits/1b4cc4ea1e4f6ffb25879df09ca9a160221b9c60
I have one question, Does callable flag means whether the issuer can early redeem the bond (does it restrict whether the investor can redeem the bond? Can the investor always early redeem regardless of the flag?

GeorgSchneider:
@judywu-da Yes, the issuer has to specify upfront if it's a callable bond (ie. issuer-redeemable). The investor can always request early/partial redemption, but the issuer is free to reject those. This is making some assumptions as to how the bond terms are defined, we might have to revisit if actual requirements are different (eg. bond can only be called on certain dates).

daml/Marketplace/Custody/Model.daml
```diff
@@ -15,29 +16,63 @@ template SettlementInfo
     operator : Party
     party : Party
     ownAccount : Optional Account
-    cashAccount : Optional Account
-    securitiesAccount : Optional Account
+    omnibusCashAccount : Optional Account
```
judywu-da, line 19: @georg-da can I regard omnibus accounts the same as settlement account or is it more like a backup account?
I can see that in Hierarchical.getAccount we will first do a lookup in cash/security accounts and if we can't find an account provide by the specify party we then look at omnibus account does this implies that omni bus account is a backup account / second choice?

daml/Marketplace/Distribution/Syndication/Investor.daml
```diff
@@ -16,15 +17,6 @@ template Role
 
     controller investor can
 
-      nonconsuming SelectSettlementAccounts : ContractId Custody.SettlementInfo
```
judywu-da: Is this removed because we will simply use the logic in Hierarchical.getAccount to find the settlement account that we need when creating instruction? (so basically, we don't give the user the right to choose which account they would like to use)

daml/Tests/Distribution/Syndication/Setup.daml
```diff
@@ -123,60 +123,62 @@ setup = do
   aiCid1 <- openAccount operator bondRegistrar bondRegistrar $ createLabel bondRegistrar bondRegistrar
   aiCid2 <- openAccount operator bondRegistrar issuer $ createLabel bondRegistrar issuer
   aiCid3 <- openAccount operator bondRegistrar custodian1 $ createLabel bondRegistrar custodian1
+  aiCid3a <- openAccount operator bondRegistrar custodian1 $ "Segregated-" <> partyToText investor1 <> "@" <> partyToText custodian1 <> "@" <> partyToText bondRegistrar
```
judywu-da, line 126: For investor1 we have this segregated account under custodian and I can see that it is set as a security account instead of omnibus account.
I guess I am confused about what account should be put as a omnibus account and what account should be insert into map of security/cash accounts.
@georg-da do you mind to clarify it for me, thanks!

daml/Marketplace/Custody/Model.daml
GeorgSchneider: The omnibus account is the account for all holdings that don't (need to) have a segregated account. So you could set up segregated accounts for some investors, and manage others in the omnibus account. Or manage all client holdings segregated, but hold your own positions in the omnibus. It's really a catch-all account that's not tied to a specific party, how this is going to be used depends on a given operating model.

daml/Tests/Distribution/Syndication/Setup.daml
GeorgSchneider: This is simply an example here that you _can_ use a segregated account for any given party, if you have to. Generally, the choice of whether to use a segregated or omnibus account depends on your operating model / regulatory / compliance constraints.

daml/Marketplace/Distribution/Syndication/Investor.daml
judywu-da: Ah I see now, thanks!

judywu-da (approved):
@dorrit-du-da changes look good to me, will add MT564 and MT566 on top of these changes.

judywu-da:
@georg-da closing this PR and remove syndication-next branch as it is replaced by syndication-dev

PR #462, opened 2022-01-31
0.1.x fix order ids

alexgraham-da:
- Allow orders to take an `Optional Int` as an order ID to let a marketplace developer to use their own order ID system instead of relying on the Exberry Adapter's orderIds
- Fix issues with the Exberry Adapter stemming from out of date Dazl functions
- Pull in some adapter/trigger tests from other branches (Partial, Cancel, MatchForever)
- Bump to `0.1.22`


brian-weir (approved):
lgtm

PR #463, opened 2022-02-01
Add nix to manage required tools for build

dasormeter:
- Add nix to manage required tools for build
  - jq
  - make
  - node
  - npm
  - pip3
  - pipenv
  - python
  - python3
  - yarn
  - yq

- Correct yq syntax for v4

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,15 @@
+# dade-installer begin
+echo "[dev-env] Setting up DA Development Environment"
+DADE_REPO_ROOT="$(pwd)"
+source /dev/stdin <<< "$(${DADE_REPO_ROOT}/dev-env/bin/dade-assist)"
+# dade-installer end
+export SBT_OPTS="-Xms2g -Xmx4g -Xss64m"
+
+export PATH="$PATH:$(pwd)/scripts"
+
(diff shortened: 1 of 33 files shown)


alexmatson-da (approved):
lgtm overall, but it could perhaps use some notes in the README on how to build the app with this

dasormeter:
> lgtm overall, but it could perhaps use some notes in the README on how to build the app with this

Shouldn't be anything special required beyond what is in the README for the building as all the tools needs get loaded / build dynamically.

But good point that we do need to say install nix before running for the first time

alexgraham-da (approved):
lgtm as well, agreed on adding to the `README`

dasormeter:
> lgtm as well, agreed on adding to the `README`

@alexgraham-da / @alexmatson-da readme updates are here https://github.com/digital-asset/da-marketplace/pull/463/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 -- can one of you try these instructions if you don't already have nix installed and see if they work?

dasormeter:
don't think we want to go all the way with nix here

PR #464, opened 2022-02-03
Add nix shell

garyverhaegen-da:
This is an alternative to #463. I'm the one who open-sourced dev-env, and I've since stopped using it, because I now believe it is simpler to use Nix directly. Before you guys make a decision, I wanted to show you what this looks like in comparison.

The main tradeoff here is that dev-env is lazy. This is good in that opening new shells and navigating to the directory remains fast regardless of whether or not you have nix tools loaded.

On the other hand, nix-shell here is eager: as soon as you cd into the directory, all of the nix tools will be built. This is good because you know you have all the dependencies so if you lose your network connection it's not an issue, and you know individual commands will not be slowed down by nix suddenly coming in and downloading stuff.

The nix-shell approach also has the advantage that you can be reasonably sure your selection is "complete" by running CI in "pure" mode where it _only_ has access to the nix-shell-provided PATH. I have not done that here, but I can point to internal examples on demand.


Makefile
```diff
@@ -1,7 +1,6 @@
-BASENAME=$(shell yq -r '.catalog.name' < dabl-meta.yaml 2> /dev/null || yq r dabl-meta.yaml 'catalog.name')
-VERSION=$(shell yq -r '.catalog.version' < dabl-meta.yaml 2> /dev/null || yq r dabl-meta.yaml 'catalog.version')
-SUBDEPLOYMENTS=$(shell yq -r '.subdeployments' < dabl-meta.yaml 2> /dev/null | sed 's/\[//g' | sed 's/\]//g' | sed 's/,//g' \
-	       || yq r dabl-meta.yaml 'subdeployments' | sed 's/\[//g' | sed 's/\]//g' | sed 's/,//g')
+BASENAME=$(shell yq '.catalog.name' dabl-meta.yaml)
+VERSION=$(shell yq '.catalog.version' dabl-meta.yaml)
+SUBDEPLOYMENTS=$(shell yq '.subdeployments[]' dabl-meta.yaml)
```
garyverhaegen-da, line 3: I'm not testing syntax for various versions, because knowing what version you're working with is the entire point of using nix to manage your tool dependencies.

nix/sources.json
```diff
@@ -0,0 +1,26 @@
+{
```
garyverhaegen-da, line 1: This is a generated file. You don't need to edit it manually.

Bump you nixpkgs definition with:

```
nix-shell -p niv --run 'niv update nixpkgs -b nixpkgs-unstable'
```
from the root of the project. It will automatically update this file for you.

nix/sources.nix
```diff
@@ -0,0 +1,174 @@
+# This file has been generated by Niv.
```
garyverhaegen-da, line 1: This is a generated file that you can completely ignore.

shell.nix
```diff
@@ -0,0 +1,20 @@
+# Update nixpkgs with:
+# nix-shell -p niv --run "niv update"
+
+let
+  sources = import ./nix/sources.nix;
+  pkgs = import sources.nixpkgs {};
+in
+pkgs.mkShell {
+  buildInputs = [
+    pkgs.bash
```
garyverhaegen-da, line 10: This is where you choose your tools. Unlike dev-env, all of the binaries provided by each package is automatically added to PATH. On the one hand, this means you don't have to make the mapping explicitly and create symlinks manually; on the other hand, it means that you can't easily alias individual binaries in case of conflicts, and that your path gets "polluted" by all the binaries of each package.

garyverhaegen-da:
I can also provide guidance for how to set this up on CircleCI, so CI works with the same set of tool versions that devs do. I haven't done that here because it's not part of #463, but would happily do it if requested.

nix/sources.json
dasormeter: @garyverhaegen-da so that is the command to regenerate this file when version upgrade is needed?

nix/sources.nix
dasormeter: this is also generated by the nix-shell package update command, or something else?

shell.nix
dasormeter: i guess same override applies if we want a specific version of the tool that does not align with the default pkgs.... version of that tool that is part of the `nixpkgs-unstable` version that we are using?

most specifically, we need particular versions of yq and nodejs/npm/yarn

nix/sources.nix
garyverhaegen-da: Also generated by `nix-shell -p niv --run "niv init -b nixpkgs-unstable"`.

You can run that in a `$(mktemp -d)` to confirm.

PR #469, opened 2022-02-10
Syndication next 2

judywu-da:
The changes in this PR is mainly for [FL-242](https://digitalasset.atlassian.net/browse/FL-242) while it also contains some changes for [FL-219](https://digitalasset.atlassian.net/browse/FL-219) and [FL-220]( https://digitalasset.atlassian.net/browse/FL-220)
Summary:

1. FL-219, FL220

> change CRDB enum to CRED/DEBT as requested by swift adaptor side

> remove assetId as input argument and return SwiftOutboundMessage contractIds for SendPrePayoutMessage choice in Lifecycle Service 

2. FL-242
> Add MT535 for security account balance reporting (aggregate by per account owner)

> Add MT940 for cash account balance reporting (aggregate by per account owner)

> Add Reporting Service for all account provider to CreateReports for all accounts under them

### DIFF
--- a/daml/SwiftMessage/Util.daml
+++ b/daml/SwiftMessage/Util.daml
@@ -15,7 +15,7 @@ data MessageType =
   | RMDR
   deriving (Eq, Show)
 
-data CRDB = CR | DB deriving (Eq, Show)
+data CRDB = CRED | DEBT deriving (Eq, Show)
 
 data PayoutDetails = PayoutDetails with
     instrumentId : Text -- ISIN + bond id label
--- a/daml/SwiftMessage/Util.daml
+++ b/daml/SwiftMessage/Util.daml
@@ -43,3 +43,9 @@ data PayoutType =
 
 getRedemptionDate : Date -> Date -> Optional Date
 getRedemptionDate maturityDate payoutDate = if payoutDate == maturityDate then Some payoutDate else Nothing
+
+data ReportDetail = ReportDetail with
+    safekeepingAccount: Account
+    assetId: Id
+    totalQuantity: Decimal
+  deriving (Eq, Show)
--- a/daml/SwiftMessage/Util.daml
+++ b/daml/SwiftMessage/Util.daml
@@ -42,7 +42,7 @@ data PayoutType =
   deriving (Eq, Show)
 
 getRedemptionDate : Date -> Date -> Optional Date
-getRedemptionDate maturityDate payoutDate = if payoutDate == maturityDate then Some payoutDate else Nothing
+getRedemptionDate maturityDate payoutDate = if payoutDate == maturityDate then Some payoutDate else None
 
 data ReportDetail = ReportDetail with
     safekeepingAccount: Account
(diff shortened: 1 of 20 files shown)


daml/SwiftMessage/Util.daml
```diff
@@ -42,4 +42,10 @@ data PayoutType =
   deriving (Eq, Show)
 
 getRedemptionDate : Date -> Date -> Optional Date
-getRedemptionDate maturityDate payoutDate = if payoutDate == maturityDate then Some payoutDate else Nothing
\ No newline at end of file
+getRedemptionDate maturityDate payoutDate = if payoutDate == maturityDate then Some payoutDate else Nothing
```
GeorgSchneider, line 45: The canonical name for the negative case is `None`, I wasn't aware that `Nothing` also exists. Or are you defining an alias somewhere?

daml/Tests/Distribution/Syndication/Lifecycle.daml
```diff
@@ -144,4 +172,11 @@ preNotify parties@Parties{..} date assetId = do
...
   snapshotCids <- map fst <$> queryFilter @Snapshot.Snapshot payingAgent (\s -> s.requester == payingAgent && s.deposit.asset.id.label == maxId.label && s.deposit.asset.id.version == maxId.version)
 
-  submit operator do exerciseByKeyCmd @Lifecycle.Service (operator, issuer, payingAgent) Lifecycle.SendPrePayoutSwiftMessage with assetId = maxId; exDivDate = date; ..
\ No newline at end of file
+  submit operator do exerciseByKeyCmd @Lifecycle.Service (operator, issuer, payingAgent) Lifecycle.SendPrePayoutSwiftMessage with exDivDate = date; ..
+  pure ()
+
+reporting : Party -> Party -> Bool -> Script ()
+reporting operator accountProvider isCashReporting = do 
+  depositCids <- map fst <$> query @AssetDeposit accountProvider
```
GeorgSchneider, line 180: You're not filtering the asset deposits here. Doesn't that mean that we also report bond holdings in the cash report?

daml/SwiftMessage/Util.daml
judywu-da: Thanks for the catch, I didn't define it anywhere else.
Not sure how does it type check.
I will update this.

daml/Tests/Distribution/Syndication/Lifecycle.daml
judywu-da: I am relying on the fact that a certain party can only be either account provider for cash or security.
Indeed, I missed out the filter
`filter (\deposit -> deposit.account.provider == accountProvider)`
as what we have in CreateReports.
Lemme add this filter.

GeorgSchneider: I think in the general case a custodian can be both cash and securities custodian. In this case they'd have to filter the deposits beforehand. I think that's acceptable for now, as we don't have a way to differentiate the two just based on the asset deposit.

daml/SwiftMessage/Util.daml
judywu-da: updated

daml/Tests/Distribution/Syndication/Lifecycle.daml
judywu-da: Noted and updated, thank you

PR #479, opened 2022-03-01
Swift Message Integration with FoP

judywu-da:
The changes in this PR includes:

1. add FOP swift messages (MT544 and MT546)
2. refactor current DvP swift messages (MT545 and MT547) along with newly added FoP messages 
3. use bond2 in the issuance script to test FoP settlement flow

Note that this change should be merged only after https://github.com/digital-asset/da-marketplace/pull/478 is merged (to avoid conflicts)


GeorgSchneider (approved):

dorrit-du-da (approved):
@judywu-da just bear in mind that if there is HTLC settlement mode in place, we need to update the changes cover in this PR.

PR #485, opened 2022-03-07
FL-296: use operator, provider and customer as actors for Terminate when possible

judywu-da:
Corresponding requirement changes described in https://digitalasset.atlassian.net/browse/FL-296?focusedCommentId=109859


dorrit-du-da (commented):
@judywu-da , would you like to make the error message explicitly? only throwing error with party names is too generic too me.

judywu-da:
@dorrit-du-da do you mean add more to the error message?
What do you think the error message should look like?

judywu-da:
updated the error messages as discussed with @dorrit-du-da

GeorgSchneider (approved):

PR #490, opened 2022-03-21
FL-326: initial draft for claim token

judywu-da:
Note that this PR serves that purpose of initial design checkpoint and will not and should not be merged into syndication-dev branch.

@georg-da I assume that the whole flow will be

1. during bond origination ClaimToken and ClaimTokenR contracts are created
2. during bond life-cycling, we will not only need to pass in AssetDeposits for specified bonds but also for corresponding claim token
3. We are not handling custodian settlement chain in TransferClaimToken (we assume that the sender and receiver accounts are provided by the same provider

### DIFF
--- a/daml/ClaimToken/Model.daml
+++ b/daml/ClaimToken/Model.daml
@@ -1,7 +1,7 @@
 module ClaimToken.Model where
 
-import DA.Finance.Types
 import DA.Finance.Asset (AssetDeposit_Transfer(..), AssetDeposit)
+import DA.Finance.Types
 import DA.Set (Set)
 
 template ClaimToken -- claim token id -> bond id
@@ -36,7 +36,6 @@ template ClaimTokenR -- reverse search (bond id -> claim token id)
 
 template TransferRelationshipOffer
   with
-    asset: Asset
     operator: Party
     provider: Party
     sender: Party
--- a/daml/ClaimToken/Model.daml
+++ b/daml/ClaimToken/Model.daml
@@ -58,7 +58,7 @@ template TransferRelationship
   where
     signatory operator, provider, sender, receiver
 
-    choice TransferClaimToken: ContractId AssetDeposit
+    nonconsuming choice TransferClaimToken: ContractId AssetDeposit
       with 
         ctrl: Party
         assetDepositCid: ContractId AssetDeposit
--- a/daml/ClaimToken/Model.daml
+++ b/daml/ClaimToken/Model.daml
@@ -19,6 +19,11 @@ template ClaimToken -- claim token id -> bond id
     ensure 
       id.signatories == originalId.signatories
 
+    choice GetOriginalId: Id
+      with ctrl: Party
+      controller ctrl 
+      do pure $ originalId
+
 template ClaimTokenR -- reverse search (bond id -> claim token id)
   with
     id: Id
@@ -33,6 +38,11 @@ template ClaimTokenR -- reverse search (bond id -> claim token id)
 
     ensure 
       id.signatories == originalId.signatories
+    
+    choice GetClaimTokenId: Id
+      with ctrl: Party
+      controller ctrl
+      do pure $ id
 
 template TransferRelationshipOffer
   with
@@ -68,6 +78,7 @@ template TransferRelationship
           assertMsg "account provider does not match" (assetDeposit.account.provider == provider && receiverAccount.provider == provider)
           assertMsg "sender should match with account owner" (assetDeposit.account.owner == sender)
           exercise assetDepositCid AssetDeposit_Transfer with receiverAccount
+          
 
 
 
--- a/daml/ClaimToken/Model.daml
+++ b/daml/ClaimToken/Model.daml
@@ -1,0 +1,76 @@
+module ClaimToken.Model where
+
+import DA.Finance.Types
+import DA.Finance.Asset (AssetDeposit_Transfer(..), AssetDeposit)
+import DA.Set (Set)
+
+template ClaimToken -- claim token id -> bond id
+  with
+    id: Id
+    originalId: Id
+    observers: Set Party
+  where
+    signatory id.signatories
+    observer observers
+
+    key (id.signatories, id.label): (Set Party, Text)
+    maintainer key._1
+
+    ensure 
+      id.signatories == originalId.signatories
+
+template ClaimTokenR -- reverse search (bond id -> claim token id)
+  with
+    id: Id
+    originalId: Id
+    observers: Set Party
+  where
+    signatory id.signatories
+    observer observers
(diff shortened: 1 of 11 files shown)


daml/ClaimToken/Model.daml
```diff
@@ -0,0 +1,76 @@
...
+    signatory id.signatories
+    observer observers
+
+    key (id.signatories, id.label): (Set Party, Text)
+    maintainer key._1
+
+    ensure 
+      id.signatories == originalId.signatories
+
+template ClaimTokenR -- reverse search (bond id -> claim token id)
```
judywu-da, line 22: @georg-da  I wonder if daml allow multiple keys on the same template. If so, we don't need ClaimTokenR.

daml/Marketplace/Settlement/Hierarchical/Util.daml
```diff
@@ -26,22 +26,26 @@ createInstructions isCash operator agent rootProvider senderAccount receiverAcco
   else if senderAccount.provider == rootProvider then do
     (_, providerSettlementInfo) <- fetchByKey @Custody.SettlementInfo (operator, receiverAccount.provider)
     let providerAccount = getAccount isCash providerSettlementInfo receiverAccount.owner
-    sis1 <- createInstruction receiverAccount.provider receiverAccount.owner (fromSome providerSettlementInfo.ownAccount) receiverAccount (asset with id.signatories = fromList [providerSettlementInfo.party]) instructionIdx
+        assetSi = if isCash then (asset with id.signatories = fromList [providerSettlementInfo.party]) else asset
```
judywu-da, line 29: This change is made so that the asset.id.signatories does not change and we can use it as key to fetch Claim Token contract for instruments that are not cash. (as bond should be fully backed by bond registrar, I think this makes sense, lemme know if otherwise)
I am not sure why we would like to change the asset.id.signatories for cash (guess something related to CBDC concept : cash provided by cash provider is different from cash provided by lead manager) @georg-da do you mind to explain a bit on this regards? thanks a lot!

daml/ClaimToken/Service.daml
```diff
@@ -0,0 +1,65 @@
...
+      observer owner
+
+      key (operator, provider, owner) : (Party, Party, Party)
+      maintainer key._1
+    
+      controller owner can 
+        nonconsuming Mint : ContractId AssetDeposit
+          with 
+            assetDepositCid : ContractId AssetDeposit
+            claimTokenR : ClaimTokenR
```
GeorgSchneider, line 22: This probably needs to be `ContractId ClaimTokenR` so that you cannot just pass any data in here.

daml/ClaimToken/Service.daml
```diff
@@ -0,0 +1,65 @@
...
+          with 
+            receiver: Party
+          do
+            create TransferRelationshipOffer with sender = owner; ..
+
+      controller owner can
+        nonconsuming Burn : ContractId AssetDeposit
+          with
+            assetDepositCid : ContractId AssetDeposit
+            claimToken : ClaimToken
```
GeorgSchneider, line 41: Same as above, needs to be `ContractId ClaimToken`

daml/Marketplace/Settlement/Hierarchical/Util.daml
GeorgSchneider: I think we need to leave that as it was. Reason being that the bond custodial holdings need to have the custodian as provider (ie. signatory), not the bond registrar. That's because they represent holdings "on behalf", ie. liabilities of the custodian, and not actual legal bond ownership (which would be at the bond registrar).

GeorgSchneider: We should ensure elsewhere that claim tokens can only be created at the top level (ie. at the bond registrar)

daml/Marketplace/Issuance/Instrument/Service.daml
```diff
@@ -27,6 +29,8 @@ template Service
           let cfi = CFI with code = "DBXXXX"
           claims <- getClaims operator provider bond.stream bond.currencyId bond.maturityDate True
           bondCid <- create bond
+          create ClaimToken with id = (bond.id with label = bond.claimTokenId); originalId = bond.id; observers = fromList [operator]
```
GeorgSchneider, line 32: Can we just pass an `Optional Text` here as the `claimTokenId` parameter, so that we don't have to store it on the `Bond` template, and we only create the claim token contracts when this is specified.

daml/ClaimToken/Service.daml
judywu-da: @GeorgSchneider this have similar issue as the HolidayCalendar that we pass in for reporting service.
ClaimToken and ClaimTokenR should be visible to public (and every party should be able to fetch them), but when the owner of the claim token are fetching it, we are not able to do something like
`controller public, owner can` (and we don't know who will own certain bond when we originate them, so we can't really put owners as observer to ClaimToken or ClaimTokenR contracts

daml/ClaimToken/Service.daml
judywu-da: @GeorgSchneider same reasoning as https://github.com/digital-asset/da-marketplace/pull/490#discussion_r836173916

daml/ClaimToken/Model.daml
judywu-da: as checked with Georg, there's no way to have multiple keys on the same template so far, will keep it as it is

daml/Marketplace/Settlement/Hierarchical/Util.daml
judywu-da: As discussed with @GeorgSchneider, since we rely mainly on Account in AssetDeposit to represent the holdings.
It's ok to keep assetDeposit.asset.id.signatories fixed on top level (ie issuer, bond registrar)

daml/ClaimToken/Service.daml
judywu-da: as discussed with @GeorgSchneider, we can use the flexible choice to return id of ClaimToken and ClaimTokenR and make those visible by public to resolve the issue here.
Updated as suggested.

daml/ClaimToken/Service.daml
judywu-da: resolved as https://github.com/digital-asset/da-marketplace/pull/490#discussion_r837065953

daml/Marketplace/Issuance/Instrument/Service.daml
judywu-da: Updated as suggested

dasormeter:
@judywu-da is this something that needs to be merged in, or can this branch be closed?

PR #492, opened 2022-03-29
Feature/fl 334 create mt 548 and integrate to issuance flow and otc trading flow

judywu-da:
The PR includes changes for [FL-334](https://digitalasset.atlassian.net/browse/FL-334):

1. add MT548 model and type
2. integrate MT548 into issuance flow (including HTLC)
3. integrate MT548 into otc-trading flow (including HTLC) 

One thing to note is that for MT548 unlike other settlement messages, we only have status `Finalized` (we don't have `Pending` status that gets converted to `Finalized` during Settle)

### DIFF
--- a/daml.yaml
+++ b/daml.yaml
@@ -24,7 +24,7 @@ sandbox-options:
   - --wall-clock-time
   - --ledgerid=da-marketplace-sandbox
 scenario-service:
-  grpc-max-message-size: 134217728
+  grpc-max-message-size: 6000000
 codegen:
   js:
     output-directory: daml.js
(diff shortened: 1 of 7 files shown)


judywu-da:
@dorrit-du-da @georg-da  you might noticed that the test is currently failing, this is due to 
`"Scenario service backend error: BErrorClient (ClientIOError (GRPCIOBadStatusCode StatusResourceExhausted (StatusDetails {unStatusDetails = \"Received message larger than max (4206356 vs. 4194304)\"})))"Script
` in Lifecycle test (basically we have too much messages after adding in MT548)
This can be easily resolved, as suggested by @georg-da after daml 2.0 upgrade with a one line daml.yaml change (I have tested locally, so we just need to wait for the upgrade branch to merge first and we can merge this PR)

dorrit-du-da (commented):
hi Judy, can we update the SettlementAdvice to SettlementStatusAndProcessingAdvice? because in SWIFT, we do have other message covering the settlmentAdvice but out of scope.

judywu-da:
> hi Judy, can we update the SettlementAdvice to SettlementStatusAndProcessingAdvice? because in SWIFT, we do have other message covering the settlmentAdvice but out of scope.

updated in the latest commit

PR #495, opened 2022-04-07
Feature/2.0.0 upgrade controller syntax

judywu-da:
controller syntax change in

1. Fin-lib dependency (syndication branch only)
2. syndication branch on current repo


GeorgSchneider (approved):
LGTM. Thanks for doing this laborious change!

PR #497, opened 2022-05-11
Feature/add par into bond

judywu-da:
The change in this PR  includes

1. add unitPar (unit to dollar ratio, e.g. 1 bond unit is 0.8 usd then unitPar = 0.8) into `Bond` contract
2. use unitPar to calculate notional value instead of using quantity for MT535 totalQuantity field (Bond Report)
3. same as https://github.com/digital-asset/da-marketplace/pull/490/files#r837065953 we will keep AssetDeposit.id.signatories same for all Bonds(non-cash assetDeposit) to easily link up asset deposit and it's corresponding `Bond` contract

### DIFF
--- a/daml/Marketplace/Issuance/Instrument/Model.daml
+++ b/daml/Marketplace/Issuance/Instrument/Model.daml
@@ -48,6 +48,7 @@ template Bond
     isPricedDirty : Bool
     isCallable : Bool
     observers : [Party]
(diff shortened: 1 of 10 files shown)


dorrit-du-da (approved):

dorrit-du-da:
as discussed offline, the par value and cash value look good

Issue #499, opened 2022-07-19
autoapproval triggers script fails with invalid input text "r"


rabadiw:
## Issue 
```bash
$ make start-autoapprove-all
/scripts/run-autoapproval-triggers.sh triggers/.daml/dist/da-marketplace-triggers-0.2.0.dar .dev
Error: 1:1: invalid input text "r"
make: *** [start-autoapprove-all] Error 1
```

## Theory
Using the pretense of `yq` version 4.26.1. The use of yq to extract the parties is not valid. 

## Solution
Change line 4 from ```parties=$(yq r daml.yaml 'parties.*')``` to ```parties=$(yq -r '.parties[]' daml.yaml)```

PR #500, opened 2022-07-19
fix: autoapproval triggers script fails with invalid input text r #499

rabadiw:
A fix for issue #499.


alexgraham-da:
I don't think it's a bad idea to update the syntax to `yq` 4.x since it is the latest version, but if you are going to do so, please update the `README.md` requirements and make sure other `yq` calls in the scripts match the new syntax as well.

PR #503, opened 2022-08-03
remove hash from canton party names

alexgraham-da:
Removes the hash from party strings in the UI as it was breaking some components.

### DIFF
--- a/ui/src/components/common/utils.ts
+++ b/ui/src/components/common/utils.ts
@@ -144,6 +144,14 @@ export function getPartyLabel(partyId: string, parties: IPartyInfo[]) {
     return { party: partyInfo?.party || partyId , label: partyInfo?.label.substring(0, partyInfo.label.lastIndexOf('|')) || partyId}
 }
 
+export const formatParty = (party: string): string => {
+  return party.split("::")[0]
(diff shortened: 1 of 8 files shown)


ui/src/components/common/utils.ts
```diff
@@ -144,6 +144,14 @@ export function getPartyLabel(partyId: string, parties: IPartyInfo[]) {
     return { party: partyInfo?.party || partyId , label: partyInfo?.label.substring(0, partyInfo.label.lastIndexOf('|')) || partyId}
 }
 
+export const formatParty = (party: string): string => {
+  return party.split("::")[0]
+}
+
+export const makePartyLabel = (party: string, name: string | undefined, role: string): string => {
```
Sarahb-da, line 151: can also be `name?: string`

Sarahb-da (approved):
lgtm

ui/src/components/common/utils.ts
alexgraham-da: I had tried that and it gave me "A required parameter cannot follow an optional parameter"...which makes sense, but I liked the order I had so kept it like this

